PR #1, opened 2022-11-28
[FLINK-30060][Connector/JDBC] Move existing JDBC connector code from Flink repo to dedicated JDBC repo

MartijnVisser:


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
```
zentol, line 3: There's a better vcs.xml example in the ES repo.

pom.xml
```diff
@@ -0,0 +1,421 @@
...
+        <testcontainers.version>1.17.2</testcontainers.version>
+        <mockito.version>2.21.0</mockito.version>
+
+        <japicmp.skip>false</japicmp.skip>
+        <japicmp.referenceVersion>1.15.0</japicmp.referenceVersion>
+
+        <slf4j.version>1.7.36</slf4j.version>
+        <log4j.version>2.17.2</log4j.version>
+
+        <test.randomization.seed/>
```
zentol, line 67: Is this actually used by any jdbc test?

pom.xml
```diff
@@ -0,0 +1,421 @@
...
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <!-- API bridge between log4j 1 and 2 -->
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-1.2-api</artifactId>
```
zentol, line 151: probably not required

pom.xml
```diff
@@ -0,0 +1,421 @@
...
+                <exclusions>
+                    <exclusion>
+                        <groupId>log4j</groupId>
+                        <artifactId>log4j</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.slf4j</groupId>
+                        <artifactId>slf4j-log4j12</artifactId>
+                    </exclusion>
+                </exclusions>
```
zentol, line 196: shouldnt be required since 1.16.0

pom.xml
```diff
@@ -0,0 +1,421 @@
...
+        <jackson-bom.version>2.13.4.20221013</jackson-bom.version>
+        <junit4.version>4.13.2</junit4.version>
+        <junit5.version>5.8.1</junit5.version>
+        <assertj.version>3.21.0</assertj.version>
+        <archunit.version>0.22.0</archunit.version>
+        <testcontainers.version>1.17.2</testcontainers.version>
+        <mockito.version>2.21.0</mockito.version>
+
+        <japicmp.skip>false</japicmp.skip>
+        <japicmp.referenceVersion>1.15.0</japicmp.referenceVersion>
```
zentol, line 62: should be 1.16.0

zentol (changes requested):
* Can also add the release_utils submodule?
* There are some references to DockerImageVersions#POSTGRES that should be replaced

pom.xml
```diff
@@ -0,0 +1,395 @@
...
+
+        <jackson-bom.version>2.13.4.20221013</jackson-bom.version>
+        <junit4.version>4.13.2</junit4.version>
+        <junit5.version>5.8.1</junit5.version>
+        <assertj.version>3.21.0</assertj.version>
+        <archunit.version>0.22.0</archunit.version>
+        <testcontainers.version>1.17.2</testcontainers.version>
+        <mockito.version>2.21.0</mockito.version>
+
+        <japicmp.skip>false</japicmp.skip>
```
zentol, line 61: We shouldn't need this here.
This is only really useful in sql-jar modules.

zentol (commented):
1 comment, otherwise +1.

zentol (approved):

PR #9, opened 2022-12-16
[FLINK-30431][Connector/JDBC] JDBC Connector fails to reestablish the lost DB connect

Jiabao-Sun:
JDBC Connector fails to reestablish the lost DB connect.

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/JdbcOutputFormat.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/JdbcOutputFormat.java
@@ -399,7 +399,14 @@ public class JdbcOutputFormat<In, JdbcIn, JdbcExec extends JdbcBatchStatementExe
     }
 
     public void updateExecutor(boolean reconnect) throws SQLException, ClassNotFoundException {
-        jdbcStatementExecutor.closeStatements();
+        try {
+            jdbcStatementExecutor.closeStatements();
+        } catch (SQLException e) {
+            // FLINK-30431 We encountered this exception may be due to the connection is closed, it
+            // depends on the behavior of different databases. To ensure that the connection can be
+            // re-established, we need to catch this exception here.
+            LOG.warn("JDBC statements were not closed gracefully.", e);
+        }
         jdbcStatementExecutor.prepareStatements(
                 reconnect
                         ? connectionProvider.reestablishConnection()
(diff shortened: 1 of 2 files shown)


MartijnVisser (commented):
@Jiabao-Sun Can this behaviour be tested?

MartijnVisser:
@Jiabao-Sun Can you rebase and use the new testing setup? @snuyanzin any thoughts on the actual PR fix itself?

snuyanzin:
I would rather go with check if preparedstatment is closed or not rather than wrapping it in `try`...`catch`.
Is there anything blocking here to go that way?

Jiabao-Sun:
> @Jiabao-Sun Can you rebase and use the new testing setup? @snuyanzin any thoughts on the actual PR fix itself?

Sorry for the late reply.
@MartijnVisser can you help trigger the CI?

Jiabao-Sun:
> I would rather go with check if preparedstatment is closed or not rather than wrapping it in `try`...`catch`. Is there anything blocking here to go that way?

Thanks @snuyanzin for the suggestion.
As the JDBC Statement definition, If the database access error occurs we may also meet SQLException. 
image

snuyanzin:
thanks for pointing to javadoc,
I've checked the javadoc for `java.sql.Statement#close` method as well e.g. [1]
and there is such statement
 >Calling the method close on a Statement object that is already closed has no effect. 

At the same time if we have a look at amazon redshift jdbc (based on trace from jira issue)
```
Caused by: com.amazon.redshift.util.RedshiftException: This connection has been closed.
    at com.amazon.redshift.jdbc.RedshiftConnectionImpl.checkClosed(RedshiftConnectionImpl.java:1095) ~[?:?]
    at com.amazon.redshift.jdbc.RedshiftConnectionImpl.cancelQuery(RedshiftConnectionImpl.java:1299) ~[?:?]
    at com.amazon.redshift.jdbc.RedshiftStatementImpl.cancel(RedshiftStatementImpl.java:1042) ~[?:?]
    at com.amazon.redshift.jdbc.RedshiftStatementImpl.close(RedshiftStatementImpl.java:748) ~[?:?]
```

it leads to `RedshiftConnectionImpl#checkClosed` [2]
```java
  protected void checkClosed() throws SQLException {
    if (isClosed()) {
      throw new RedshiftException(GT.tr("This statement has been closed."),
          RedshiftState.OBJECT_NOT_IN_STATE);
    }
  }
```

it violates the contracts mentioned in javadoc....
I would incline to consider that this is a redshift jdbc driver bug
[1] https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#close--
[2] https://github.com/aws/amazon-redshift-jdbc-driver/blob/master/src/main/java/com/amazon/redshift/jdbc/RedshiftStatementImpl.java#L828-L833

Jiabao-Sun:
Thanks @snuyanzin to point out this problem.
I think it is indeed the driver that violates this specification.
But back to this issue, do we need to fix it by capturing or wait for the redshift driver to fix this problem?

snuyanzin:
short answer: i would suggest to wait for driver's fix

a bit more about what make me thinking like that
1. since there are multiple drivers might work with jdbc connector and each could have it's own issues it does not make sense to work around such issues on connectors level
2. there are lots of methods which could throw `SQLException` and at the same time handling of just one/part of them looks not consistent 
3. such wrapping in `try`...`catch` could potentially hide some other `SQLException` (not related to the current one) which for some other reasons could appear... Thus it could make it complex to analyze issue in case of any

PR #15, opened 2023-01-05
[FLINK-30576] Refactor JdbcOutputFormat to use ExecutionConfig

eskabetxe:
To be able to use the JdbcOutputFormat on the new sink based on SinkV2, we need to change the JdbcOutputFormat to be free of RuntimeContext

the JdbcOutputFormat need to know if ObjectReuse is active.

Some notes:
- should we change the name of JdbcOutputFormat, as it is no longer extending OutputFormat?
- the JdbcOutputFormat.Builder is not used internally, should we remove the builder or force the creation of JdbcOutputFormat only in the builder?


wanglijie95:
@eskabetxe Can we start reviewing this change after  the FLIP is approved ? Because we don’t know what we can get at the moment, executionConfig or objectReuse, or something else, which will decide how we modify the `OutputFormat#open`.

eskabetxe:
closing this PR.. only 1 PR will be submitted with all changes

PR #20, opened 2023-02-03
[FLINK-29750][Connector/JDBC]Support to write data to jdbc when disable flush-interval and max-rows

hehuiyuan:

sink.buffer-flush.interval | The flush interval mills, over this time, asynchronous threads will flush data. Can be set to '0' to disable it. Note, 'sink.buffer-flush.max-rows' can be set to '0' with the flush interval set allowing for complete async processing of buffered actions.
-- | --
sink.buffer-flush.max-rows | The max size of buffered records before flush. Can be set to zero to disable it.

`Asynchronous thread flush` is disabled&nbsp; when set `sink.buffer-flush.interval` = 0,&nbsp; then disable `sink.buffer-flush.max-rows`, which causes the data is not written out.

The data are not written when disable `flush.interval` and `max-rows`. I think it should write out when the data arrive.

image


hehuiyuan:
@MartijnVisser hi， moved here.

The flink pr address:
https://github.com/apache/flink/pull/19297

eskabetxe:
@hehuiyuan I think this is expected...
Setting 'sink.buffer-flush.max-rows' will make that the sink only write records at checkpoint.
If you want to write records as they arrive, you should set it to 1.

eskabetxe:
@MartijnVisser for me is the expected behavior, so yes we can close it.

PR #21, opened 2023-02-08
[FLINK-30963][ci] Streamline binary URL configuration

zentol:
The ci_utils workflow can now automatically infer the download URL and caching behavior. This implicitly switches stable binary downloads to archive.apache.org. This link will never expire; while the download speed is worse it shouldn't be a problem due to caching.

https://github.com/zentol/flink-connector-jdbc/actions/runs/4123496195/jobs/7121586447


MartijnVisser (approved):
Nice :)

PR #23, opened 2023-02-16
[hotfix][doc] Fix the description about lookup.max-retries in the zh docs

Jiabao-Sun:
The config option `lookup.max-retries` should be translated as 重试次数。


Jiabao-Sun:
Hi @leonardBang.
Could you help review this?

leonardBang (approved):
Thanks @Jiabao-Sun for the fix, LGTM

PR #24, opened 2023-02-22
[FLINK-31181] Support LIKE operator pushdown

grzegorz8:


libenchao (approved):
+1, @grzegorz8 Thanks for your contribution!

PR #26, opened 2023-03-10
[Improve]Increase the valid of judging connection.

lightzhao:


lightzhao:
@kezhuw @leonardBang @wanglijie95 PTAL.

PR #28, opened 2023-03-17
[hotfix] [doc] Fix the configuration item error of docs/content.zh/docs/connectors/table/jdbc.md

xuzhiwen1255:
Fix lookup.cache.max-rows in docs/content.zh/docs/connectors/table/jdbc.md configuration

## Brief change log
Fix option(lookup.cache.max-rows -> lookup.partial-cache.max-rows ) in jdbc.md

Dependencies (does it add or upgrade a dependency): no
The public API, i.e., is any changed class annotated with @public(Evolving): no
The serializers: no
The runtime per-record code paths (performance sensitive): no
Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: no
The S3 file system connector: no

## Documentation
Does this pull request introduce a new feature? no

### DIFF
--- a/docs/content.zh/docs/connectors/table/jdbc.md
+++ b/docs/content.zh/docs/connectors/table/jdbc.md
@@ -210,7 +210,7 @@ ON myTopic.key = MyUserTable.id;
       <td>维表的缓存策略。 目前支持 NONE（不缓存）和 PARTIAL（只在外部数据库中查找数据时缓存）。</td>
     </tr>
     <tr>
-      <td><h5>lookup.cache.max-rows</h5></td>
+      <td><h5>lookup.partial-cache.max-rows</h5></td>
       <td>可选</td>
       <td style="word-wrap: break-word;">(none)</td>
       <td>Integer</td>


MartijnVisser (changes requested):
@xuzhiwen1255 For me its not clear why this is an error. The English docs state for this config option:

`The max number of rows of lookup cache, over this value, the oldest rows will be expired. Lookup cache is disabled by default. See the following Lookup Cache section for more details.`

That seems correct?

xuzhiwen1255:
@MartijnVisser  His description is correct, but his configuration key is incorrectly named.
Because he's using the old key.

But when I look at the code,jdbc collector does not use the latest cacheOptions, so it should not be changed. Do we need to replace the obsolete options with the latest options?

xuzhiwen1255:
@MartijnVisser Sorry, i made a mistake, it should be that the current jdbc lookup configuration is obsolete, and we should replace it with the latest lookup configuration.

eg. Configure ``lookup.cache.max-rows`` to replace ``lookup.partial-cache.max-rows`` ..

Otherwise, when users configure according to the official website information, they are still using the old configuration parameters instead of the latest ones.
I think we should modify all obsolete parameters and replace them with the latest lookup configuration.

What do you think? @MartijnVisser

PR #31, opened 2023-03-28
[FLINK-31649] Fix archunit violations

eskabetxe:
This PR fix the archunit violations from
- ITCASE tests should use a MiniCluster resource or extension
- Return and argument types of methods annotated with @PublicEvolving must be annotated with @Public(Evolving)


flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
```diff
@@ -28,6 +30,7 @@
 
 /** Thread-safe (assuming immutable {@link Xid} implementation). */
 @ThreadSafe
+@PublicEvolving
```
wanglijie95, line 33: How about removing the `@PublicEvolving` of `XaSinkStateHandler`? Because it seems that it should not be a user interface (only used in `JdbcXaSinkFunction`, and `JdbcXaSinkFunction` is `@Internal`).
Once `XaSinkStateHandler` is not `@PublicEvolving`, the change here will be not needed.

cc @rkhachatryan

wanglijie95 (commented):
Thanks for creating this PR @eskabetxe. I left a comment, PTAL. Besides, you missed the `archunit-violations/b10c194a-0517-4744-830f-81abfb71ab18`

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
eskabetxe: Hi @wanglijie95,
Its a param of the constructor of JdbcXaSinkFunction (line 199), because of that I set it to PublicEvolving.
All params should be public no?

eskabetxe:
> Thanks for creating this PR @eskabetxe. I left a comment, PTAL. Besides, you missed the `archunit-violations/b10c194a-0517-4744-830f-81abfb71ab18`

No, I left it as I really dont know how to solve this..
this org.apache.flink.connector.jdbc.internal.options.JdbcConnectorOptions is annotated with Internal and in internal package, but extends org.apache.flink.connector.jdbc.JdbcConnectionOptions that is PublicEvolving

then we have another org.apache.flink.connector.jdbc.table.JdbcConnectorOptions that is PublicEvolving and extends nothing.

I think the first must be set as Public, but we will need to move to another package.
Maybe we can mix the two JdbcConnectorOptions only into one..

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
wanglijie95: But the `JdbcXaSinkFunction` is `@Internal`, which means the users are not expectd to use it directly(call any constructor of it)

eskabetxe: yes true, changed

wanglijie95:
> org.apache.flink.connector.jdbc.internal.options.JdbcConnectorOptions

This is also `@Internal`, which is only used by `JdbcDynamicTableSink`(users are not expected to use it directly). It breaks the rule because it ends with `ConnectorOptions` (see `TableApiRules#CONNECTOR_OPTIONS_PACKAGE` for details).

I think a simple solutation is  rename it(to `JdbcOptions` or `JdbcConnectorInternalOptions` or some other names, not ending with `ConnectorOptions` is fine.)

eskabetxe:
Renamed, and rebased

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
wanglijie95, line 33: The `JdbcXaSinkFunctionState` should not mark as `PublicEvolving` due to `XaSinkStateHandler` is `@Internal` now. We should change it to `@Internal`

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
```diff
@@ -53,7 +56,7 @@ private JdbcXaSinkFunctionState(
      * @return immutable collection of prepared XA transactions to {@link
      *     javax.transaction.xa.XAResource#commit commit}.
      */
-    public Collection<CheckpointAndXid> getPrepared() {
+    protected Collection<CheckpointAndXid> getPrepared() {
```
wanglijie95, line 59: This change is not needed now.

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/options/JdbcConnectionOptions.java
```diff
@@ -32,15 +31,15 @@
 
 /** Options for the JDBC connector. */
 @Internal
-public class JdbcConnectorOptions extends JdbcConnectionOptions {
+public class JdbcConnectionOptions extends org.apache.flink.connector.jdbc.JdbcConnectionOptions {
```
wanglijie95, line 34: It would be better to change this class to another name, not same as the existing `JdbcConnectionOptions`

wanglijie95 (commented):
@eskabetxe I left some comments, PTAL.

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
eskabetxe: done

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
eskabetxe: reverted

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/options/JdbcConnectionOptions.java
eskabetxe: Changed

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/xa/JdbcXaSinkFunctionState.java
wanglijie95: It seems it hasn't been updated yet?

eskabetxe: sorry, updated..

wanglijie95 (approved):
Thanks @eskabetxe . LGTM.

PR #35, opened 2023-04-03
[hotfix][docs] Fix the broken postgresql driver download link

hackergin:
Fix the broken postgresql driver download link


MartijnVisser (approved):
@hackergin Thanks, LGTM

PR #38, opened 2023-04-10
[FLINK-31759][JDBC] Update some external connectors to use sql_connector_download_table and connector_artifact shortcode

reswqa:


TanYuxin-tyx (approved):
@reswqa Thanks for fixing the doc version. LGTM

PR #39, opened 2023-04-13
[FLINK-31699][Connectors/JDBC] Fixing dependency convergence

dannycranmer:
Fixes https://github.com/apache/flink-connector-jdbc/actions/runs/4687686153/jobs/8307285788

Tested via
- `mvn clean install -Pcheck-convergence -Dflink.version=1.16.1`
- `mvn clean install -Pcheck-convergence -Dflink.version=1.17.0`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@ under the License.
 
     <properties>
         <flink.version>1.16.0</flink.version>
-        <flink.shaded.version>15.0</flink.shaded.version>
+        <flink.shaded.version>16.1</flink.shaded.version>
 
         <jackson-bom.version>2.13.4.20221013</jackson-bom.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -48,7 +48,7 @@ under the License.
 
     <properties>
         <flink.version>1.16.0</flink.version>
-        <flink.shaded.version>15.0</flink.shaded.version>
+        <flink.shaded.version>16.1</flink.shaded.version>
```
MartijnVisser, line 51: We actually should create a follow-up ticket because connectors shouldn't rely on Flink Shaded at all, like we've also done for Cassandra https://issues.apache.org/jira/browse/FLINK-31722

MartijnVisser (approved):
LGTM

PR #44, opened 2023-04-27
Fix duplicated data on retrying

ZhaoxuanSu:
When SQL execution fails, Flink will retry. At this point, the PreparedStatement still contains the previously added parameters, resulting in data duplication.

Should we always clear previously added parameters on 'executeBatch' method to avoid this?


MartijnVisser (changes requested):
@ZhaoxuanSu Please have a look at https://flink.apache.org/how-to-contribute/overview/ on how to contribute code; there should be Jira ticket for this issue, and it needs to be properly linked to this PR too

PR #45, opened 2023-05-04
[hotfix] Fix incorrect sql_url in jdbc.yml.

reswqa:


docs/data/jdbc.yml
```diff
@@ -18,4 +18,4 @@
 
 variants:
   - maven: flink-connector-jdbc
-    sql_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-sql-connector-jdbc/$full_version/flink-sql-connector-jdbc-$full_version.jar
+    sql_url: https://repo.maven.apache.org/maven2/org/apache/flink/flink-connector-jdbc/$full_version/flink-connector-jdbc-$full_version.jar
```
reswqa, line 21: It should be noted that in the flink-1.17 document, this link will be replaced with `https://repo.maven.apache.org/maven2/org/apache/flink/flink-connector-jdbc/3.0.0-1.17/flink-connector-jdbc-3.0.0-1.17.jar` and is currently inaccessible. After the jdbc connector releases a version that supports 1.17, the URL will be automatically activated.

reswqa:
@leonardBang would you mind taking a look at this? Thanks~

docs/data/jdbc.yml
leonardBang: shouldn't `3.1.0-1.17`?

reswqa: IIRC, `v3.1.0` should not have been released yet. So, I guess you mean that the jdbc connector supports `flink-1.17` will be the version `v3.1.0`.

At present, the `release-1.17` branch of flink tracks the `v3.0` branch of the `flink-connnector-jdbc` repository. 
https://github.com/apache/flink-connector-jdbc/blob/b07bddfc8d741e203d3d9b6eeefe5113ecfb4acc/docs/content/docs/connectors/table/jdbc.md?plain=1#L41
This line indicates that the connector version is `v3.0.0`, which will cause `$full_version` to be replaced with `3.0.0-1,17`. If we want to release `v3.1.0`, modifying the version number here should be a necessary part of the release process.

leonardBang: > At present, the `release-1.17` branch of flink tracks the `v3.0` branch of the `flink-connnector-jdbc` repository.

IIUC, `3.0.0-1.17` never released and looks we don't have plan to release  this version, https://mvnrepository.com/artifact/org.apache.flink/flink-connector-jdbc

reswqa: If `3.1.0` is the only version that support `flink-1.17`, there are two possibilities when releasing:

- Create `v3.1` branch: let the `setup_docs.sh` tracks this branch.

- Do not create `v3.1` branch: modify the doc in `jdbc-connector` repository, let `sql_connector_download_table "jdbc"` point to `3.1.0`.

Both of these will automatically change the download url linked to `3.1.0-1.17`.

leonardBang (approved):
LGTM

reswqa:
Thanks for the review, merged.

PR #48, opened 2023-05-11
[BP-3.0][FLINK-32024][docs] Short code related to externalized connector retr…

reswqa:
…ieve version from its own data yaml

### DIFF
--- a/docs/data/jdbc.yml
+++ b/docs/data/jdbc.yml
@@ -16,6 +16,7 @@
 # limitations under the License.
 ################################################################################
 
+version: 3.1.0
 variants:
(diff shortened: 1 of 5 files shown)


docs/data/jdbc.yml
```diff
@@ -16,6 +16,7 @@
 # limitations under the License.
 ################################################################################
 
+version: 3.1.0
```
TanYuxin-tyx, line 19: LGTM for the version fix, but the test failed, we should fix that.

TanYuxin-tyx (approved):

PR #51, opened 2023-05-31
[FLINK-32228] Bump testcontainers version

eskabetxe:
Bump testcontainers version to last one


eskabetxe:
@MartijnVisser could you check?

pom.xml
```diff
@@ -53,7 +53,7 @@ under the License.
         <junit4.version>4.13.2</junit4.version>
         <junit5.version>5.9.1</junit5.version>
         <assertj.version>3.23.1</assertj.version>
-        <testcontainers.version>1.17.2</testcontainers.version>
+        <testcontainers.version>1.18.1</testcontainers.version>
```
MartijnVisser, line 56: Any particular reason why we don't go straight for 1.18.2?

MartijnVisser (approved):
LGTM, just curious if we shouldn't upgrade to one more newer version

pom.xml
eskabetxe: I check here https://mvnrepository.com/artifact/org.testcontainers/testcontainers
and that was the latest version..

I will update to 1.18.2

snuyanzin (approved):

pom.xml
eskabetxe: updated

PR #54, opened 2023-06-08
'jsonb' postgres type doesn't support yet

m1karmida:
Hi, i need to use the connector for interacting with a postgres database. When i tried to retrieve a table with a 'jsonb' type column,   an exeption with the message in title occured. Temporarily, i cloned the project and patched it adding in the datatype mapper the case 'jsonb', but this solution is not portable because i used maven dependency, and this one matches only locally.


eskabetxe:
Hi @m1karmida, please have a look on [how to contribute code](https://flink.apache.org/how-to-contribute/overview/); there should be Jira ticket for this issue, and it needs to be properly linked to this PR too

And you should create your branch from master and do the PR to master (not from 3.1 to master)..
You can ask to back-ported this to 3.1 after

MartijnVisser:
This can't be completed without https://issues.apache.org/jira/browse/FLINK-29267 being done. Closing until that's resolved.

PR #63, opened 2023-06-16
Update AbstractPostgresCompatibleDialect.java

xushsh163:
When there is a conflict between a field name and a keyword in a PostgreSQL table, it is necessary to use quotation marks

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
@@ -69,7 +69,13 @@ public abstract class AbstractPostgresCompatibleDialect extends AbstractDialect 
 
     @Override
     public String quoteIdentifier(String identifier) {
-        return identifier;
+    	if (identifier.contains(".")) {
+    		String[] ids = identifier.split("\\.");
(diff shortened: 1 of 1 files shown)


MartijnVisser (changes requested):
@xushsh163 Thanks for the PR, can you please make sure that a test is added for this scenario. Ideally with a test we would be able to identify that this is indeed not working without your change

xushsh163:
@MartijnVisser 
In postgresql 13.10
`CREATE TABLE TEST(
   "LEFT" CHAR(20)
);`
In Flink 1.16.0 SQL-Client
`CREATE TABLE test
(
  `LEFT` STRING
) WITH (
  'connector' = 'jdbc',
  'url' = 'jdbc:postgresql://xxxxx:5432/reporting',
  'table-name' = 'test',
  'username' = 'xxx',
  'password' = 'xxx'
);

select * from test limit 10;`

Error:
[ERROR] Could not execute SQL statement. Reason:
org.postgresql.util.PSQLException: ERROR: syntax error at or near "FROM"
  Position: 13

MartijnVisser:
> In postgresql 13.10

Yes, but I would like to see that added as a test in this PR, so it can be automatically included and validated in the future as well.

xushsh163:
@MartijnVisser Sorry, I am not a professional open-source maintainer, so it is difficult for me to write a standardized test case that meets your project's specifications. Perhaps you can treat this PR as an issue and fix it by your team. Then I will close this PR.

PR #64, opened 2023-06-25
Fixed missing words in documentation

baxinyu:
对flink-connector-jdbc/docs/content.zh/docs/connectors/table/jdbc.md中60行的错词漏字进行修正


baxinyu:
The flink - connector - JDBC/docs/content. useful/docs/connectors/table/JDBC md pragraphs 60 lines in the wrong word

PR #68, opened 2023-07-03
Jsonb support

m1karmida:
I created this PR for supporting the JSONB postgres object in the catalog. The PR simply performs a conversion in a string, so in this way it is possible to leverage all JSON functions provided by the official documentation.


MartijnVisser:
This requires a proper FLIP and implementation, see also https://issues.apache.org/jira/browse/FLINK-29267. Next to that, please follow the contributing code process as documented at https://flink.apache.org/how-to-contribute/contribute-code/

PR #74, opened 2023-09-11
[FLINK-33069]Mysql and Postgres catalog support url extra parameters

waywtdcc:
Mysql and Postgres catalog support url extra parameters

CREATE CATALOG mymysql WITH(
'type' = 'jdbc',
'default-database' = 'test',
'username' = 'root',
'password' = 'xxx',
'base-url' = 'jdbc:mysql://xxx:53309',
'extra-url-param' = '?characterEncoding=utf8'
);

If used in this way, the URLs of all tables obtained from this catalog are: jdbc:mysql://xxx:53309?characterEncoding=utf8

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/catalog/AbstractJdbcCatalog.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/catalog/AbstractJdbcCatalog.java
@@ -92,6 +92,7 @@ public abstract class AbstractJdbcCatalog extends AbstractCatalog {
     protected final String pwd;
     protected final String baseUrl;
     protected final String defaultUrl;
+    protected final String extraUrlParam;
 
     public AbstractJdbcCatalog(
             ClassLoader userClassLoader,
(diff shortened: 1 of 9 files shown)


waywtdcc:
@MartijnVisser Hello, can you review this PR when you have time?

MartijnVisser:
@waywtdcc Why does this need to be an additional parameter?

waywtdcc:
> @waywtdcc Why does this need to be an additional parameter?

Because sometimes you need to use ?characterEncoding=utf8 similar url parameters to connect to jdbc

MartijnVisser:
> Because sometimes you need to use ?characterEncoding=utf8 similar url parameters to connect to jdbc

But why not just pass these in to the base URL?

waywtdcc:
> > Because sometimes you need to use ?characterEncoding=utf8 similar url parameters to connect to jdbc
> 
> But why not just pass these in to the base URL?

The real url is a combination of base-url+default-database, so I cannot form the correct url when I add it to base-url.

waywtdcc:
> > Because sometimes you need to use ?characterEncoding=utf8 similar url parameters to connect to jdbc
> 
> But why not just pass these in to the base URL?

The real url is a combination of base-url+default-database, so I cannot form the correct url when I add it to base-url. For example, if the base-url is set to jdbc:mysql://xxx:53309?characterEncoding=utf8, then the final real url is jdbc:mysql://xxx:53309?characterEncoding=utf8/test; this is the wrong url.

PR #82, opened 2023-12-08
[FLINK-33787][jdbc] Java 17 support for jdbc connector

snuyanzin:
After FLINK-33302 is finished it is now possible to specify jdk version
That allows to add jdk17 support

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -34,6 +34,7 @@ jobs:
           branch: main
         }, {
           flink: 1.18-SNAPSHOT,
+          jdk: '8, 11, 17',
           branch: main
         }, {
           flink: 1.16.2,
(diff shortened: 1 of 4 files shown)


.github/workflows/weekly.yml
```diff
@@ -45,6 +45,14 @@ jobs:
           flink: 1.18.0,
           branch: v3.1
         }]
+        jdk: [ 8, 11 ]
+        include:
+          - flink: 1.18-SNAPSHOT
+            branch: main
+            jdk: 17
```
eskabetxe, line 52: cant we add this jdk inside the the flink_branches on the 1.18 options..
Like overriding the default... or is not allowed?

snuyanzin: the issue is that since shared utill `ci_utils` does not have versions and that means it will be automatilly applied for all connectors and will make all of them failing since most of them do not support java 17

eskabetxe: ok..
I was thinking something like this:
`matrix:
        flink_branches: [
        ....,{
          flink: 1.17.1,
          branch: v3.1
        }, {
          flink: 1.18.0,
          branch: v3.1,
          jdk: [ 8, 11, 17 ]`
        }]
        jdk: [ 8, 11 ]`

snuyanzin: seems didn't get it first, sorry

yes, that's possible the only diff that it should be string, otherwise sequence in nested fields are not allowed
Moreover strings allow to get the ci job view in a nicer way

snuyanzin:
looks like gha fails to download flink, will retry a bit later

eskabetxe (approved):
LGTM

PR #83, opened 2023-12-11
[FLINK-33800][JDBC/Connector] Allow passing parameters to database via jdbc url

snuyanzin:
The idea is pretty straightforward: from one side there is a jdbc url which is normally is provided by most of the dbs providers to connect to. 
From the other side for some,  probably historical reasons, there a separate property for default database which is used for default url.

To not break former behaviour it is possible to specify either database in jdbcurl or database with extra parameters only if database is same as ifor a dedicated property for default database. 
In fact I tend to think to mark it as a deprecated since all the info could be extracted from jdbc url


eskabetxe:
Hi @snuyanzin, 
I can imagine that the intention would be to have all the db's with a single catalog, but being mandatory the default-database my assumption can be wrong..

Saying that and being one catalog for db, won't be better to add a new option like url (as table have), and deprecate this two.
It becomes a little strange to have the db on the two option, and base-url being a full url..

eskabetxe:
this is also related with this: https://github.com/apache/flink-connector-jdbc/pull/74

PR #87, opened 2023-12-28
[FLINK-33929][jdbc-connector] Support JDBC String field read Fragment read

zhilinli123:
https://issues.apache.org/jira/browse/FLINK-33929?filter=-2

### DIFF
--- a/docs/content.zh/docs/connectors/datastream/jdbc.md
+++ b/docs/content.zh/docs/connectors/datastream/jdbc.md
@@ -84,6 +84,60 @@ env.from_collection(
 
 env.execute()
 ```
+### 支持字符串分片读取
+<table class="table table-bordered" style="width: 60%">
+    <thead>
+      <tr>
+        <th class="text-left">Support String Split Database</th>
+        <th class="text-left">Slice Read String Grammar</th>
+       </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>MySQL</td>
+            <td>ABS(MD5(`column`) % `number_of_partitions`)</td>
+        </tr>
+        <tr>
+            <td>Oracle</td>
+            <td>MOD(ORA_HASH(`column`) , `number_of_partitions`)</td>
+        <tr>
+            <td>PostgreSQL</td>
+            <td>(ABS(HASHTEXT(`column`)) % `number_of_partitions`)</td>
+        </tr>
+        <tr>
+            <td>MS SQL Server</td>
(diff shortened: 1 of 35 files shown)


zhilinli123:
PTAL: @MartijnVisser thanks !

MartijnVisser:
@zhilinli123 I still don't understand the Jira ticket, so I can't really review it. Perhaps @snuyanzin or @eskabetxe understand the goal of this PR.

zhilinli123:
> @zhilinli123 I still don't understand the Jira ticket, so I can't really review it. Perhaps @snuyanzin or @eskabetxe understand the goal of this PR.

hi~ @MartijnVisser  The current jdbc slicing support type is limited, currently does not support for string type fields for slicing read, many table design using UUID as the table key without self-increment primary key, resulting in the use of flink reading is no way to slice read, the current function is to support the current shortage

eskabetxe:
> @zhilinli123 I still don't understand the Jira ticket, so I can't really review it. Perhaps @snuyanzin or @eskabetxe understand the goal of this PR.

For what I can assume, @zhilinli123 are trying to use scan.partition on a table that is hash partitioned by a string field..

My understanding of scan.partition is that only divides the query on multiples query based on params of scan.partition but is not using partitioning of table, it only adds "field BETWEEN ? AND ?" in query, and for that it only allow numeric, date, or timestamp column..

zhilinli123:
> > @zhilinli123 I still don't understand the Jira ticket, so I can't really review it. Perhaps @snuyanzin or @eskabetxe understand the goal of this PR.
> 
> For what I can assume, @zhilinli123 are trying to use scan.partition on a table that is hash partitioned by a string field..
> 
> My understanding of scan.partition is that only divides the query on multiples query based on params of scan.partition but is not using partitioning of table, it only adds "field BETWEEN ? AND ?" in query, and for that it only allow numeric, date, or timestamp column..

You're right, but slicing string hash md5 doesn't need to be done BETWEEN? AND ?  It's much more efficient to check this shard directly

zhilinli123:
PTALA: @libenchao  hi, could you please help review this function? I would like to express my heartfelt thanks

zhilinli123:
PTAL:@eskabetxe @leonardBang @snuyanzin If you have time, could you review it for me? Express gratitude

zhilinli123:
If you have time, could you review it for me? Express gratitude @MartijnVisser @eskabetxe

PR #88, opened 2024-01-04
[FLINK-33970] Add necessary checks for connector document

GOODBOY008:
Add necessary checks for connector document to avoid flink document build failure.


GOODBOY008:
@MartijnVisser Please take a look, Thank you~

MartijnVisser (changes requested):
@GOODBOY008 Like I've said in the Jira, this shouldn't be addressed per connector but generically for the shared_utils.

Next to that, I'm not sure this solves the problem. You're only checking for dead links in the existing connector docs, but it doesn't actually check the integration of the docs when the Flink documentation is being built. That's where the error occurred.

eskabetxe (approved):
LGTM

GOODBOY008:
@Jiabao-Sun PTAL

GOODBOY008:
@MartijnVisser PTAL , Thank you~

PR #89, opened 2024-01-16
[FLINK-34088] [hotfix] fix the problems with special table name characters of postgres and oracle  and sqlserver.

BlackPigHe:
[hotfix] fix the problems with special table name characters of postgres and oracle  and sqlserver.

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
@@ -168,4 +168,14 @@ public class JdbcUtils {
         }
         return pkRow;
     }
+
+    public static String handleDoubleQuotes(String identifier) {
+        String[] split = identifier.split("\\.");
+        StringBuilder builder = new StringBuilder();
+        for (String s : split) {
+            builder.append("\"").append(s).append("\"");
+            builder.append(".");
+        }
+        return builder.deleteCharAt(builder.length() - 1).toString();
+    }
 }
--- a/flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/databases/oracle/dialect/OraclePreparedStatementTest.java
+++ b/flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/databases/oracle/dialect/OraclePreparedStatementTest.java
@@ -48,10 +48,10 @@ class OraclePreparedStatementTest {
         String insertStmt = dialect.getInsertIntoStatement(tableName, fieldNames);
         assertThat(insertStmt)
                 .isEqualTo(
-                        "INSERT INTO tbl(id, name, email, ts, field1, field_2, __field_3__) "
+                        "INSERT INTO \"tbl\"(\"id\", \"name\", \"email\", \"ts\", \"field1\", \"field_2\", \"__field_3__\") "
                                 + "VALUES (:id, :name, :email, :ts, :field1, :field_2, :__field_3__)");
         NamedStatementMatcher.parsedSql(
-                        "INSERT INTO tbl(id, name, email, ts, field1, field_2, __field_3__) "
+                        "INSERT INTO \"tbl\"(\"id\", \"name\", \"email\", \"ts\", \"field1\", \"field_2\", \"__field_3__\") "
                                 + "VALUES (?, ?, ?, ?, ?, ?, ?)")
                 .parameter("id", singletonList(1))
                 .parameter("name", singletonList(2))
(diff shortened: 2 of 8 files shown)


BlackPigHe:
add the commit to update test case

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
```diff
@@ -168,4 +168,14 @@ public static Row getPrimaryKey(Row row, int[] pkFields) {
...
+
+    public static String handleDoubleQuotes(String identifier) {
+        String[] split = identifier.split("\\.");
+        StringBuilder builder = new StringBuilder();
+        for (String s : split) {
+            builder.append("\"").append(s).append("\"");
+            builder.append(".");
+        }
+        return builder.deleteCharAt(builder.length() - 1).toString();
+    }
```
snuyanzin, line 180: I didn't get why we are going to put logic with double quotes into some generic place if for different engines there could be different quotes?

flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/databases/oracle/dialect/OraclePreparedStatementTest.java
```diff
@@ -48,10 +48,10 @@ void testInsertStatement() {
         String insertStmt = dialect.getInsertIntoStatement(tableName, fieldNames);
         assertThat(insertStmt)
                 .isEqualTo(
-                        "INSERT INTO tbl(id, name, email, ts, field1, field_2, __field_3__) "
+                        "INSERT INTO \"tbl\"(\"id\", \"name\", \"email\", \"ts\", \"field1\", \"field_2\", \"__field_3__\") "
                                 + "VALUES (:id, :name, :email, :ts, :field1, :field_2, :__field_3__)");
         NamedStatementMatcher.parsedSql(
-                        "INSERT INTO tbl(id, name, email, ts, field1, field_2, __field_3__) "
+                        "INSERT INTO \"tbl\"(\"id\", \"name\", \"email\", \"ts\", \"field1\", \"field_2\", \"__field_3__\") "
```
snuyanzin, line 54: In issue description it was only something about table name
why does it impact columns?

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
BlackPigHe: Because oracle and postgres have the same escape characters, It's all double quotes

flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/databases/oracle/dialect/OraclePreparedStatementTest.java
BlackPigHe: the method ‘quoteIdentifier’ is common，column metadata is also by call the method.Columns should have this problem as well

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
snuyanzin: this package is not only for oracle and postgres. E.g. MS SQL Server, MySQL are different

snuyanzin:
I guess we can extract information about current quote identifier directly from JDBCDriver 
`java.sql.DatabaseMetaData#getIdentifierQuoteString`
and then we don't need to hardcode it for every dialect

flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/databases/oracle/dialect/OraclePreparedStatementTest.java
snuyanzin: can we have a a filaing test reproducing the problem?

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/utils/JdbcUtils.java
BlackPigHe: Do you think that the location of this method is not appropriate, I think it is OK, the specific escape characters of the data source in their own package implementation, there are public features to extract and did not find a better location

BlackPigHe:
> I guess we can extract information about current quote identifier directly from JDBCDriver `java.sql.DatabaseMetaData#getIdentifierQuoteString` and then we don't need to hardcode it for every dialect

In this case, special characters of various data sources need to be preset, and the workload is relatively large

BlackPigHe:
> I guess we can extract information about current quote identifier directly from JDBCDriver `java.sql.DatabaseMetaData#getIdentifierQuoteString` and then we don't need to hardcode it for every dialect

I get it. If there are no compatibility issues, It's better this way.

BlackPigHe:
> > I guess we can extract information about current quote identifier directly from JDBCDriver `java.sql.DatabaseMetaData#getIdentifierQuoteString` and then we don't need to hardcode it for every dialect
> 
> I get it. If there are no compatibility issues, It's better this way.
I just had a look at it and found that the interface was in the test package

BlackPigHe:
> quote identifier

I thought the method `Java.SQL.DatabaseMetaData#getIdentifierQuoteString` to dynamically access quote identifier, the price is too big, need to introduce Connection object which is very heavy, It is not reasonable for the existing dialect class design.

snuyanzin:
yep, may be it is a point to see whether we could adapt it or not 

btw I haven't found any test reproducing the issue, can you add one?
Also I would say the issue is more complex than just adding sql quote identifiers
e.g. even though there are quotes queries with fields containing question marks `?` will fail FLINK-34211
i guess we need to a have a number of dedicated tests for that

PR #91, opened 2024-01-17
[FLINK-34128] [bugfix] Some jdbc objects cannot be obtained properly in oracle jdbc,Specify the required type explicitly

BlackPigHe:


snuyanzin:
Thanks for the contribution
could you please add a test case reproducing the original issue and confirming that the PR fixes it?

BlackPigHe:
@snuyanzin Building this test case requires using jdbc to link to oracle, which can be difficult to simulate.Can I simulate with mock data. haha

BlackPigHe:
Finally found that this is an old version of the bug, when I only looked at the location of the error and found that the new code did not change this area, I raised an issue, in fact, this problem has already been fixed in another place

PR #93, opened 2024-01-18
[hotfix] Add jakarta.activation required after changes in Flink main repo

snuyanzin:
The reason is this line https://github.com/apache/flink/blame/a41229b24d82e8c561350c42d8a98dfb865c3f69/flink-runtime/src/test/java/org/apache/flink/runtime/io/network/partition/hybrid/tiered/netty/TestingTierConsumerAgent.java#L28
added within https://github.com/apache/flink/pull/23927

now we need to have this dependency for tests to continue build it with jdk17+

P.S. to reproduce problem locally 
```
mvn clean install -Dflink.version=1.19-SNAPSHOT
```
(assuming that it will be run with jdk17 or jdk21 + assuming that either there is no builds in local maven cache of Flink 1.19-SNAPSHOT or there is a build for latest commit)

### DIFF
--- a/flink-connector-jdbc/pom.xml
+++ b/flink-connector-jdbc/pom.xml
@@ -248,7 +248,12 @@ under the License.
 			<artifactId>flink-architecture-tests-test</artifactId>
 			<scope>test</scope>
 		</dependency>
-
+		<dependency>
+			<groupId>jakarta.activation</groupId>
+			<artifactId>jakarta.activation-api</artifactId>
+			<version>1.2.1</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<dependencyManagement>


BlackPigHe:
<dependency>
			<groupId>javax.activation</groupId>
			<artifactId>activation</artifactId>
			<version>1.1.1</version>
			<scope>test</scope>
		</dependency>
增加这个包就可以啦

BlackPigHe:
I also found this problem, and found that the work flow kept failing when mr

snuyanzin:
@BlackPigHe i didn't get your comment
have you seen the changes proposed within this PR?

MartijnVisser (approved):
LGTM. Should we file a Jira and do this for more connectors?

snuyanzin:
Thanks, for having a look
however i'm now wondering whether we can just wait and close it later since it's going to be fixed in Flink itself
https://github.com/apache/flink/pull/24147

snuyanzin:
closing since the issue was fixed in upstream

PR #98, opened 2024-01-30
[FLINK-34259] Fix fails to compile with NPE on hasGenericTypesDisabled

Jiabao-Sun:
[FLINK-34259] Fix fails to compile with NPE on hasGenericTypesDisabled

### DIFF
--- a/flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/JdbcTestBase.java
+++ b/flink-connector-jdbc/src/test/java/org/apache/flink/connector/jdbc/JdbcTestBase.java
@@ -59,9 +59,13 @@ public abstract class JdbcTestBase implements DerbyTestBase {
         return context;
     }
 
-    public static ExecutionConfig getExecutionConfig(Boolean reused) {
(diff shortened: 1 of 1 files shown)


Jiabao-Sun:
CI passed locally https://github.com/Jiabao-Sun/flink-connector-jdbc/actions/runs/7709389700/job/21010504922

MartijnVisser (changes requested):
@Jiabao-Sun This shouldn't be fixed in the JDBC connector, this appears to be a blocker for 1.19 since they shouldn't break interfaces

Jiabao-Sun:
> IIUC, we're just failing the test not compiling, right?

Yes

snuyanzin:
@reswqa can you please elaborate why it was decided to merge it here nevertheless it was mentioned by @MartijnVisser 
>@Jiabao-Sun This shouldn't be fixed in the JDBC connector, this appears to be a blocker for 1.19 since they shouldn't break interfaces

I feel like we need more info

snuyanzin:
ok, thanks for clarification
i should have looked at jira as well

PR #99, opened 2024-01-31
[hotfix] Use 1.16.x as minimum supported version in pom

snuyanzin:
As it was identified at https://github.com/apache/flink-web/pull/707#discussion_r1472769071
there should be minimum Flink version in pom and in push pr as well

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@ under the License.
     </modules>
 
     <properties>
-        <flink.version>1.17.0</flink.version>
+        <flink.version>1.16.3</flink.version>
 
         <jackson-bom.version>2.13.4.20221013</jackson-bom.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -47,7 +47,7 @@ under the License.
     </modules>
 
     <properties>
-        <flink.version>1.17.0</flink.version>
+        <flink.version>1.16.3</flink.version>
```
MartijnVisser, line 50: We could even set it to 1.16.0 (since that's the lowest supported version), even though there shouldn't be any changes in a patch version of Flink

MartijnVisser (approved):

pom.xml
snuyanzin: I changed to 1.16.0, thanks
will merge after ci becomes green

PR #100, opened 2024-02-02
[FLINK-34014][jdbc] Avoid executeBatch when buffer is empty

jeyhunkarimov:
This PR adds a few checks to avoid sending empty buffer to database

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/executor/TableBufferedStatementExecutor.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/executor/TableBufferedStatementExecutor.java
@@ -52,11 +52,13 @@ public final class TableBufferedStatementExecutor implements JdbcBatchStatementE
 
     @Override
     public void executeBatch() throws SQLException {
-        for (RowData value : buffer) {
-            statementExecutor.addToBatch(value);
+        if (!buffer.isEmpty()) {
+            for (RowData value : buffer) {
+                statementExecutor.addToBatch(value);
+            }
+            statementExecutor.executeBatch();
+            buffer.clear();
         }
-        statementExecutor.executeBatch();
-        buffer.clear();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


jeyhunkarimov:
Hi @snuyanzin could you please review this PR in your available time? Thanks!

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/executor/TableBufferedStatementExecutor.java
```diff
@@ -52,11 +52,13 @@ public void addToBatch(RowData record) throws SQLException {
...
+        if (!buffer.isEmpty()) {
+            for (RowData value : buffer) {
+                statementExecutor.addToBatch(value);
+            }
+            statementExecutor.executeBatch();
+            buffer.clear();
         }
-        statementExecutor.executeBatch();
-        buffer.clear();
     }
```
morazow, line 62: Would it make sense to return early if buffer is empty in both methods? (It would show the change clearly)

affo: +1

affo (commented):
Looks quite good, apart that I would also prefer as @morazow to see an early `return` instead of a big `if` block.

Just a note, isn't there any test to change because of this?
Also nothing to add to verify that no 'blank' call is made?

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/executor/TableBufferedStatementExecutor.java
snuyanzin: From one side yes from another side current change is done in a way of already existing style like e.g. similar existing code for 
`org.apache.flink.connector.jdbc.internal.executor.InsertOrUpdateJdbcExecutor#executeBatch`
may be we need to refactor all of them at some point

snuyanzin (approved):

PR #108, opened 2024-03-21
[FLINK-34901][connectors/jdbc]Improve the performance of the update Postgres database

Mrart:
Improve the performance of the update Postgres database

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
@@ -56,6 +56,7 @@ public abstract class AbstractPostgresCompatibleDialect extends AbstractDialect 
                         .collect(Collectors.joining(", "));
         String updateClause =
                 Arrays.stream(fieldNames)
+                        .filter(f -> !Arrays.asList(uniqueKeyFields).contains(f))
                         .map(f -> quoteIdentifier(f) + "=EXCLUDED." + quoteIdentifier(f))
                         .collect(Collectors.joining(", "));
         return Optional.of(


Mrart:
@1996fanrui Can you help me review?

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
```diff
@@ -56,6 +56,7 @@ public Optional<String> getUpsertStatement(
                         .collect(Collectors.joining(", "));
         String updateClause =
                 Arrays.stream(fieldNames)
+                        .filter(f -> !Arrays.asList(uniqueKeyFields).contains(f))
```
1996fanrui, line 59: Would you mind elaborate on why adding this filter?

If adding this filter makes sense, I left 2 comments about code logic:
- We can initialize the `Arrays.asList(uniqueKeyFields)` only once. IIUC, this array will be converted to List multiple times when `fieldNames` length > 0.
- We call the contains method, it's better to use Set instead of List. right?

1996fanrui (commented):
Thanks @Mrart for the contribution. I left some comments on the code change.

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
Mrart: Thanks @1996fanrui , I think that was a little bit inaccurate, update clause must exclude unique index . If not will case execptions like 'modification of distribution columns in OnConflictUpdate is not supported Call getNextException to see other errors in the batch'.

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
```diff
@@ -50,12 +50,14 @@ public String getLimitClause(long limit) {
...
             String tableName, String[] fieldNames, String[] uniqueKeyFields) {
+        Set<String> uniqueKeyFieldsSet = Arrays.stream(uniqueKeyFields).collect(Collectors.toSet());
         String uniqueColumns =
-                Arrays.stream(uniqueKeyFields)
+                uniqueKeyFieldsSet.stream()
                         .map(this::quoteIdentifier)
                         .collect(Collectors.joining(", "));
         String updateClause =
                 Arrays.stream(fieldNames)
+                        .filter(f -> !uniqueKeyFieldsSet.contains(f))
```
1996fanrui, line 60: I'm curious is this a bug or performance improvement?

Also cc @RocMarshal , would you mind reviewing this PR as well (as you are the active contributor of jdbc connector)?

1996fanrui:
Hi @Mrart , the CI is failed, could you rebase the main branch and check the CI?

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/dialect/AbstractPostgresCompatibleDialect.java
1996fanrui: It's better to add a test to check it if it's a bug.

RocMarshal: Thanks @Mrart for the contribution & @1996fanrui for the ping.

> It's better to add a test to check it if it's a bug.

+1.

PR #112, opened 2024-04-08
[FLINK-34961] Use dedicated CI name for JDBC connector to differentiate it in infra-reports

snuyanzin:
The PR will allow to differentiate between jdbc connector statistics and others with name ci


RyanSkraba (approved):
LGTM

PR #115, opened 2024-04-18
[FLINK-33460][Connector/JDBC] Support property authentication connection.

RocMarshal:
- Support property authentication connection.

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
@@ -98,13 +98,6 @@ public class JdbcConnectionOptions implements Serializable {
             return this;
         }
 
-        public JdbcConnectionOptionsBuilder withProperties(Properties properties) {
-            if (Objects.nonNull(properties)) {
-                this.properties.putAll(properties);
-            }
-            return this;
-        }
-
         public JdbcConnectionOptionsBuilder withProperty(String propKey, String propVal) {
             Preconditions.checkNotNull(propKey, "Connection property key mustn't be null");
             Preconditions.checkNotNull(propVal, "Connection property value mustn't be null");
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
@@ -20,35 +20,39 @@ package org.apache.flink.connector.jdbc;
 import org.apache.flink.annotation.PublicEvolving;
 import org.apache.flink.util.Preconditions;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import java.io.Serializable;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Properties;
 
 /** JDBC connection options. */
 @PublicEvolving
 public class JdbcConnectionOptions implements Serializable {
+
+    public static final String USER_KEY = "user";
+    public static final String PASSWORD_KEY = "password";
 
     private static final long serialVersionUID = 1L;
 
     protected final String url;
     @Nullable protected final String driverName;
     protected final int connectionCheckTimeoutSeconds;
-    @Nullable protected final String username;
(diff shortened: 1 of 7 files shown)


RocMarshal:
hi, @snuyanzin @eskabetxe Could you help to have a review if you had the free time ?
Thank you~

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
```diff
@@ -90,6 +101,11 @@ public JdbcConnectionOptionsBuilder withDriverName(String driverName) {
             return this;
         }
 
+        public JdbcConnectionOptionsBuilder withExtendProperties(Properties extendProperties) {
```
eskabetxe, line 104: won't be better have the method to allow add one property 

```
private Properties properties = new Properties();
public JdbcConnectionOptionsBuilder withConfig(String config, String value) {
            this.properties.put(config, value);
            return this;
}
```

this way we could also start thinking on moving user, pass and driver to this, as they will be used always inside the properties..
simplifying some logic in [SimpleJdbcConnectionProvider.java](https://github.com/apache/flink-connector-jdbc/pull/115/files#diff-1a261021e67f437344c5b6b2b5da274ff6914c51d8b5163ef30f03e6ba3614c5)

```
public JdbcConnectionOptionsBuilder withUsername(String username) {
            this.properties.put("user", username);
            return this;
}
```

eskabetxe (commented):
@RocMarshal left one comment as as I think that will help to simplify later some code..

overall LGTM..

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
```diff
@@ -90,13 +98,31 @@ public JdbcConnectionOptionsBuilder withDriverName(String driverName) {
             return this;
         }
 
+        public JdbcConnectionOptionsBuilder withProperties(Properties properties) {
```
eskabetxe, line 101: I have mixed feelings about this method...

I understand that it may be practical, but as a user I can expect that the properties I am passing will override all properties already defined.

as an example:
```
.withProperty("a", "a1")
.withProperties(new properties())
.withProperty("b", "b1")
```

If you saw that (without knowing the implementation), what will you expect.
that a and b should be in the properties, or just b...

eskabetxe (approved):
LGTM

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/JdbcConnectionOptions.java
RocMarshal: hi, @eskabetxe Thanks a lot for your quick-review 😄 . 

```
.withProperty("a", "a1")
.withProperties(new properties())
.withProperty("b", "b1")
``` 

> .....
> that a and b should be in the properties, or just b...
It's right.
It's difficult to express , which depends the what cases devs need and I have to admit that the lines referenced above is ambiguous now.  

I have come up with two temporary ways to alleviate this ambiguity:

- Keep the current logic, and then we add comment to describe semantics for these two methods.

OR

- Add a flag parameter to method `withProperties(new properties(), true/false)` , indicating whether to force assign   entire `properties`, rather than a part of the `properties` by `upsert` mode.

Of course, I'd appreciated to hear any ideas form you / community devs ! 😃

eskabetxe: @RocMarshal I would say the easiest way, eliminating the method.

If someone has a properties, they can do:
`properties.forEach((key, value) -> builder.withProperty(key, value));`

Under the hood, the `properties.putAll` is what it does.

RocMarshal: Thanks for the comments.
That sounds nice to me 
updated~

RocMarshal:
hi, @Jiabao-Sun master, Could you help to have a review if you had the free time ?
Thank you~ 😄

1996fanrui (approved):
Thanks @RocMarshal for the contribution, and @eskabetxe for the review!

LGTM

PR #121, opened 2024-05-14
[FLINK-35333] Remove weekly check for v3.1 with 1.19

eskabetxe:
the branch cut for v3.1 was not prepared for 1.19 versions

in another PR we introduce this check, but I think that should be avoided as the cut is not prepared for that..

to prepare the branch to 1.19 we will have to bring some changes that are in master to solve some 1.19 problems


eskabetxe:
@MartijnVisser I think this is the best solution

snuyanzin (approved):

PR #130, opened 2024-06-06
[FLINK-35542]  Set the correct class loader to lookup for the class CheckpointedOffset

jan-osc:
- [FLINK-35542]  Use the class loader of class CheckpointedOffset while deserializing JdbcSourceSplit.

### DIFF
--- a/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/source/split/JdbcSourceSplitSerializer.java
+++ b/flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/source/split/JdbcSourceSplitSerializer.java
@@ -106,7 +106,8 @@ public class JdbcSourceSplitSerializer implements SimpleVersionedSerializer<Jdbc
         byte[] chkOffsetBytes = new byte[chkOffsetBytesLen];
         in.read(chkOffsetBytes);
         CheckpointedOffset chkOffset =
-                InstantiationUtil.deserializeObject(chkOffsetBytes, CheckpointedOffset.class.getClassLoader());
+                InstantiationUtil.deserializeObject(
+                        chkOffsetBytes, CheckpointedOffset.class.getClassLoader());
 
         return new JdbcSourceSplit(id, sqlTemplate, params, offset, chkOffset);
     }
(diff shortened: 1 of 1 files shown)


flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/source/split/JdbcSourceSplitSerializer.java
```diff
@@ -106,7 +106,7 @@ public JdbcSourceSplit deserializeJdbcSourceSplit(DataInputStream in)
         byte[] chkOffsetBytes = new byte[chkOffsetBytesLen];
         in.read(chkOffsetBytes);
         CheckpointedOffset chkOffset =
-                InstantiationUtil.deserializeObject(chkOffsetBytes, in.getClass().getClassLoader());
+                InstantiationUtil.deserializeObject(chkOffsetBytes, CheckpointedOffset.class.getClassLoader());
```
RocMarshal, line 109: Thanks @jan-osc for the pr.
Would you like to add a test case for the fix ?

jan-osc: Hello. This bug only appears when the JDBC connector is deployed to the cluster during the deserialization of the `JdbcSourceSplit` from the state. I'll think through how I could simulate these conditions in tests and if it's really needed. In the simplest test case, we could just test if serialization/deserialization works but with a JUnit class loader (flat class loader). Class loader setup in Flink is more complicated (FlinkUserCodeClassLoader).

RocMarshal (approved):
@jan-osc Thanks a lot for the pr.
LGTM +1.
CC @snuyanzin @1996fanrui

1996fanrui (approved):
Thanks for the fix!

LGTM

flink-connector-jdbc/src/main/java/org/apache/flink/connector/jdbc/source/split/JdbcSourceSplitSerializer.java
1996fanrui, line 109:
```suggestion
                InstantiationUtil.deserializeObject(
                        chkOffsetBytes, CheckpointedOffset.class.getClassLoader());
```

PR #134, opened 2024-07-29
[FLINK-35909][test] Using reflection to resolve constructor conflicts

whhe:


flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/table/source/JdbcFilterPushdownPreparedStatementVisitorTest.java
```diff
@@ -294,4 +294,40 @@ private List<ResolvedExpression> resolveSQLFilterToExpression(
...
+
+            Constructor<?>[] constructors = clazz.getConstructors();
+
+            for (Constructor<?> constructor : constructors) {
+                if (constructor.getParameterCount() == 4) {
+                    return (RexNodeToExpressionConverter)
+                            constructor.newInstance(
+                                    rexBuilder, inputNames, functionCatalog, catalogManager);
+                }
+                if (constructor.getParameterCount() == 5) {
```
eskabetxe, line 318:
```suggestion
                if (constructor.getParameterCount() == 5 &&
                        TimeZone.class.equals(constructor.getParameters()[4].getType())) {
```

eskabetxe (approved):
LGTM, @1996fanrui can you check?

I would prefer not to use reflection to solve a backward compatibility problem in same Flink version, but since it’s just a test, I can't raise too many issues.

1996fanrui (approved):
Thanks for the fix and review!

LGTM

1996fanrui:
The CI of main branch is green for now, thanks again.

PR #67, opened 2023-07-03
[FLINK-30702] Add Elasticsearch dialect

grzegorz8:
https://issues.apache.org/jira/browse/FLINK-30702

PR #136, opened 2024-08-06
Add Elasticsearch dialect

grzegorz8:
The PR adds Elasticsearch dialect.

Important facts:
 * `x-pack-sql-jdbc` dependency is on Elastic license which is not compatible with Apache license. Therefore:
     * `x-pack-sql-jdbc` has `provided` scope.
     * appropriate note is added in the documentation.
     * added an enforcer rule to make sure that it's never bundled.
 * Elastic jdbc driver is read-only.
     * In integration tests the test indices and test data are created via Elastic REST API.

---
This PR replaces https://github.com/apache/flink-connector-jdbc/pull/67. Due to recent refactoring it was easier to create a new PR.

### DIFF
--- a/docs/content.zh/docs/connectors/table/jdbc.md
+++ b/docs/content.zh/docs/connectors/table/jdbc.md
@@ -47,17 +47,18 @@ JDBC 连接器不是二进制发行版的一部分，请查阅[这里]({{< ref "
 
 在连接到具体数据库时，也需要对应的驱动依赖，目前支持的驱动如下：
 
-| Driver     | Group Id                   | Artifact Id            | JAR                                                                                                                         |
-|:-----------|:---------------------------|:-----------------------|:----------------------------------------------------------------------------------------------------------------------------|
-| MySQL      | `mysql`                    | `mysql-connector-java` | [下载](https://repo.maven.apache.org/maven2/mysql/mysql-connector-java/)                                                      |
-| Oracle     | `com.oracle.database.jdbc` | `ojdbc8`               | [下载](https://mvnrepository.com/artifact/com.oracle.database.jdbc/ojdbc8)                                                    | 
-| PostgreSQL | `org.postgresql`           | `postgresql`           | [下载](https://jdbc.postgresql.org/download/)                                                                                 |
-| Derby      | `org.apache.derby`         | `derby`                | [下载](http://db.apache.org/derby/derby_downloads.html)                                                                       |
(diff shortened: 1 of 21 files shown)


eskabetxe (commented):
Overall, everything seems to be fine. 
However, I suggest changing the module name to "flink-connector-jdbc-elasticsearch."

I still have some doubts about:
- whether this should be part of the Elasticsearch connector: "https://github.com/apache/flink-connector-elasticsearch."
- the use of trial version

The recent refactor makes this feasible, as you only need to import jdbc-core.

@1996fanrui, what do you think?
@RocMarshal, could you take a look as well?

grzegorz8:
> I still have some doubts about:
>   * whether this should be part of the Elasticsearch connector: "https://github.com/apache/flink-connector-elasticsearch."
> The recent refactor makes this feasible, as you only need to import jdbc-core.
> 
> @1996fanrui, what do you think? @RocMarshal, could you take a look as well?

Hi! @1996fanrui @RocMarshal Any thoughts?

1996fanrui:
Sorry, I'm not familiar with jdbc and elasticserarch connector.

flink-connector-jdbc-elasticsearch/pom.xml
```diff
@@ -0,0 +1,122 @@
...
+        </dependency>
+
+
+        <!-- Elasticsearch -->
+        <dependency>
+            <groupId>org.elasticsearch.plugin</groupId>
+            <artifactId>x-pack-sql-jdbc</artifactId>
+            <version>${elasticsearch.version}</version>
+            <scope>provided</scope>
+        </dependency>
```
linghengqian, line 73: - According to https://discuss.elastic.co/t/the-documentation-does-not-seem-to-document-the-minimum-jdk-runtime-requirement-for-org-elasticsearch-plugin-x-pack-sql-jdbc/372078/3?u=linghengqian , the committer of ElasticSearch pointed out that `org.elasticsearch.plugin:x-pack-sql-jdbc` is mainly used inside ElasticSearch. It is not suitable for scenarios including Flink.

> I don't understand. The plugin runs inside Elasticsearch so it uses the JVM provided by Elasticsearch.
> It's not meant to be reused anywhere else.

dadoonet: No sorry. I was wrong for this jar.

PR #143, opened 2024-10-09
[FLINK-36009] Architecture tests refactor

eskabetxe:
Created the module [flink-connector-jdbc-architecture](https://issues.apache.org/jira/browse/FLINK-connector-jdbc-architecture) to have all tests, we need to add the modules as dependency to validate that module..


RocMarshal (commented):
Hi, @eskabetxe Thank you for your hard work.
LGTM on the whole just the commit message header.
image

Would you mind helping update it ? thx~

eskabetxe:
> Hi, @eskabetxe Thank you for your hard work. LGTM on the whole just the commit message header. image
> 
> Would you mind helping update it ? thx~

fixed, the incorrect one was the PR that was using the Epic

eskabetxe:
@1996fanrui can you check

1996fanrui (commented):
> @1996fanrui can you check

This PR is related to `Architecture refactor`, since I'm not an expert of jdbc-connector, so it's better to involve a commiter or PMC member related to connector, thanks.

cc @leonardBang

eskabetxe:
@leonardBang @MartijnVisser can you help here..

flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/CompatiblePostgresDialectConverter.java
```diff
@@ -18,11 +18,11 @@
 
 package org.apache.flink.connector.jdbc.postgres.database.dialect;
 
-import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.annotation.Internal;
 import org.apache.flink.table.types.logical.RowType;
 
 /** JDBC converter for PostgreSQL compatible databases. */
-@PublicEvolving
+@Internal
```
snuyanzin, line 25: this is unclear to me
shouldn't we create a kind of FLIP first if we make such changes?

eskabetxe: This configuration is currently misaligned. All dialect implementations are classified as internal. The purpose of this class is solely to allow the CrateDB dialect to utilize the PostgreSQL implementation.

To preserve the PublicEvolving status, we need to add an exception to the architecture test. Without this exception, the rule "Public Evolving API Methods Use Only Public Evolving API Types" will fail.

To resolve this issue without an exception, we would need to refactor all dialect implementations and classify them as PublicEvolving.

I can add the exception if you still believe that a FLIP is necessary and focus the discussion solely on that change. I believe this will not impact the rest of the PR.

snuyanzin: >I can add the exception if you still believe that a FLIP is necessary and focus the discussion solely on that change. I believe this will not impact the rest of the PR.

would be great.
Agree, rest of the PR is more or less ok (I did only a brief look, will look again a bit later)

eskabetxe: @snuyanzin already changed and added the exception to architecture tests.
When you have time could you check again.. 
Thanks

snuyanzin (approved):

PR #148, opened 2024-11-28
[FLINK-36816] Support source parallelism setting for JDBC connector

grzegorz8:
# Purpose of the change

Add new option `scan.parallelism` support for JDBC connector.

Part of [FLINK-33261](https://issues.apache.org/jira/browse/FLINK-33261) FLIP-367: Support Setting Parallelism for Table/SQL Sources

The change is not backward compatible for versions <= 1.18. `InputFormatProvider of(InputFormat<RowData, ?> inputFormat, @Nullable Integer sourceParallelism)` was added in Flink 1.19. Therefore, we need to wait with this change until JDBC connector is no longer supporting Flink 1.18.

### DIFF
--- a/docs/content.zh/docs/connectors/table/jdbc.md
+++ b/docs/content.zh/docs/connectors/table/jdbc.md
@@ -213,6 +213,14 @@ ON myTopic.key = MyUserTable.id;
       <a href="https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor">Postgres</a>，可能需要将此设置为 false 以便流化结果。</td>
     </tr>
     <tr>
+      <td><h5>scan.parallelism</h5></td>
+      <td>optional</td>
+      <td>no</td>
+      <td style="word-wrap: break-word;">(none)</td>
+      <td>Integer</td>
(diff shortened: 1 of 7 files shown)


docs/content.zh/docs/connectors/table/jdbc.md
```diff
@@ -212,6 +212,14 @@ ON myTopic.key = MyUserTable.id;
       它决定了每个语句是否在事务中自动提交。有些 JDBC 驱动程序，特别是
       <a href="https://jdbc.postgresql.org/documentation/head/query.html#query-with-cursor">Postgres</a>，可能需要将此设置为 false 以便流化结果。</td>
     </tr>
+    <tr>
+      <td><h5>scan.parallelism</h5></td>
```
davidradl, line 216: could you add a link to some documentation for this - as there is for autocommit please.

RocMarshal (commented):
Thanks @grzegorz8 for the contribution .
Could you still work for this patch ?

If there is no response for an extended period, we will try to find other developers to continue advancing this patch. BTW, your initial contribution will be retained as a co-author in the commit message.

Looking forward to your back or reply!

PR #150, opened 2024-12-17
[FLINK-36236] JdbcSourceBuilder overrides set connection provider

coopstah13:
Only instantiate the provider from the connection options builder if no provider was set on the builder


eskabetxe:
Thanks for the improvement @coopstah13 
Could you add some test to validate this. 
Thanks

ferenc-csaky:
Superseded by #169

PR #155, opened 2025-02-05
[hotfix][Connector/JDBC] Remove unnecessary unique fields of PostgreSQL upsert statement

roseduan:
In PostgreSQL, there is no need to specify the unique key fields in the `DO UPDATE SET` clause.

See more details in https://www.postgresql.org/docs/current/sql-insert.html

### DIFF
--- a/flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresDialect.java
+++ b/flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresDialect.java
@@ -25,6 +25,7 @@ import org.apache.flink.table.types.logical.RowType;
 
 import java.util.Arrays;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresDialect.java
```diff
@@ -73,8 +74,10 @@ public Optional<String> getUpsertStatement(
                 Arrays.stream(uniqueKeyFields)
                         .map(this::quoteIdentifier)
                         .collect(Collectors.joining(", "));
+        final Set<String> uniqueKeyFieldsSet = new HashSet<>(Arrays.asList(uniqueKeyFields));
         String updateClause =
                 Arrays.stream(fieldNames)
+                        .filter(f -> !uniqueKeyFieldsSet.contains(f))
```
leonardBang, line 80: This change makes sense to me after read Postgresql `ON CONFLICT ... DO UPDATE SET..` SQL grammar, could you add a unit test for this change ?

roseduan: Thanks for review. I have added a unit test.

flink-connector-jdbc-postgres/src/test/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresPreparedStatementTest.java
```diff
@@ -0,0 +1,46 @@
...
+    private final JdbcDialect dialect =
+            JdbcFactoryLoader.loadDialect(
+                    "jdbc:postgresql://localhost:3306/test", getClass().getClassLoader());
+    private final String[] fieldNames =
+            new String[] {"id", "name", "email", "ts", "field1", "field_2", "__field_3__"};
+    private final String[] keyFields = new String[] {"id", "__field_3__"};
+    private final String tableName = "tbl";
+
+    @Test
+    void testUpsertStatement() {
```
leonardBang, line 40: We can add this test case to `PostgresDialectTest`

flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresDialect.java
```diff
@@ -73,17 +74,22 @@ public Optional<String> getUpsertStatement(
...
                         .map(f -> quoteIdentifier(f) + "=EXCLUDED." + quoteIdentifier(f))
                         .collect(Collectors.joining(", "));
+        String conflictAction =
+                updateClause.isEmpty()
+                        ? " DO NOTHING"
+                        : String.format(" DO UPDATE SET %s", updateClause);
```
leonardBang, line 86: Could we add two IT cases to cover the two situations?

roseduan: OK, fixed.

leonardBang (approved):
Thanks @roseduan for the update, LGTM

PR #156, opened 2025-02-09
[FLINK-37288] Add Google Cloud Spanner dialect and catalog

laughingman7743:

Adds Google Cloud Spanner support to flink-connector-jdbc as a new `flink-connector-jdbc-spanner` module.

- **Dialect**: Type mapping between Spanner and Flink types, including array type write support
- **Catalog**: Database/table discovery via Spanner Admin API and `INFORMATION_SCHEMA`, with semicolon-delimited URL parameter handling (`;param=value`)
- **Lineage**: Extracts Spanner-specific location info (project, instance, database) from JDBC URLs
- **Docs**: Added Spanner section to JDBC connector docs (English and Chinese)

### Changes to core module

- `AbstractJdbcCatalog`: Added protected constructor and hooks (`calculateUrlFunction`, `validateConnectionProperties`) for dialects with non-standard URL formats
- `FieldNamedPreparedStatement`: Array type support

## Testing

All features are tested against the Spanner emulator via Testcontainers: dialect type mapping, catalog operations, end-to-end read/write, and factory SPI registration.

## Jira

https://issues.apache.org/jira/browse/FLINK-37288

## Discussion

https://lists.apache.org/thread/4344z501pszv5h7gmr0311fw7h6h4sw4


flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
```diff
@@ -125,17 +125,10 @@ public AbstractJdbcCatalog(
...
         this.userClassLoader = userClassLoader;
+        this.connectionProperties = Preconditions.checkNotNull(connectionProperties);
         this.baseUrl = baseUrl.endsWith("/") ? baseUrl : baseUrl + "/";
         this.defaultUrl = getDatabaseUrl(defaultDatabase);
-        this.connectionProperties = Preconditions.checkNotNull(connectionProperties);
-        checkArgument(
-                !StringUtils.isNullOrWhitespaceOnly(connectionProperties.getProperty(USER_KEY)));
-        checkArgument(
-                !StringUtils.isNullOrWhitespaceOnly(
-                        connectionProperties.getProperty(PASSWORD_KEY)));
```
laughingman7743, line 138: Spanner does not use password authentication.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
```diff
@@ -576,16 +569,6 @@ protected String getSchemaTableName(ObjectPath tablePath) {
...
-    /**
-     * URL has to be without database, like "jdbc:dialect://localhost:1234/" or
-     * "jdbc:dialect://localhost:1234" rather than "jdbc:dialect://localhost:1234/db".
-     */
-    protected static void validateJdbcUrl(String url) {
-        String[] parts = url.trim().split("\\/+");
-
-        checkArgument(parts.length == 2);
-    }
-
```
laughingman7743, line 588: In the case of Spanner, the URL is as follows, so I have deleted this validation.
```
jdbc:cloudspanner://hostname/projects/gcp_project_id/instances/instance_id/databases/database_id
```

flink-connector-jdbc-spanner/src/test/java/org/apache/flink/connector/jdbc/spanner/SpannerTestBase.java
```diff
@@ -16,27 +16,19 @@
...
+import org.apache.flink.connector.jdbc.testutils.DatabaseTest;
+import org.junit.jupiter.api.extension.ExtendWith;
 
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
+/** Base class for Spanner testing. */
+@ExtendWith(SpannerDatabase.class)
+public interface SpannerTestBase extends DatabaseTest {
 
-/** Test for {@link AbstractJdbcCatalog}. */
-class AbstractJdbcCatalogTest {
```
laughingman7743, line 26: This class is no longer needed because URL validation has been removed.

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/testutils/tables/TableBase.java
```diff
@@ -120,7 +122,7 @@ public int[] getTableTypes() {
                 .toArray();
     }
 
-    public Schema getTableSchema() {
+    public Schema getTableSchema(String pkConstraintName) {
```
laughingman7743, line 125: In the case of Spanner, the contract name of the primary key is different, so it is possible to specify the contract name of the primary key.

laughingman7743:
I have applied code formatting using Spotless and updated the documentation. It is ready for review.

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -45,17 +45,18 @@ See how to link with it for cluster execution [here]({{< ref "docs/dev/configura
...
+| CrateDB    | `io.crate`                 | `crate-jdbc`                | [Download](https://repo1.maven.org/maven2/io/crate/crate-jdbc/)                                                                         |
+| Db2        | `com.ibm.db2.jcc`          | `db2jcc`                    | [Download](https://www.ibm.com/support/pages/download-db2-fix-packs-version-db2-linux-unix-and-windows)                                 | 
+| Trino      | `io.trino`                 | `trino-jdbc`                | [Download](https://repo1.maven.org/maven2/io/trino/trino-jdbc/)                                                                         |
+| OceanBase  | `com.oceanbase`            | `oceanbase-client`          | [Download](https://repo1.maven.org/maven2/com/oceanbase/oceanbase-client/)                                                              |
+| Spanner    | `com.google.cloud`         | `google-cloud-spanner-jdbc` | [Download](https://central.sonatype.com/artifact/com.google.cloud/google-cloud-spanner-jdbc)                                            |
```
davidradl, line 59: I see the download url points to Maven like the spanner doc. How do I get the jar file?

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -702,7 +708,7 @@ SELECT * FROM given_database.test_table2;
 
 Data Type Mapping
 ----------------
-Flink supports connect to several databases which uses dialect like MySQL, Oracle, PostgreSQL, CrateDB, Derby, SQL Server, Db2 and OceanBase. The Derby dialect usually used for testing purpose. The field data type mappings from relational databases data types to Flink SQL data types are listed in the following table, the mapping table can help define JDBC table in Flink easily.
+Flink supports connect to several databases which uses dialect like MySQL, Oracle, PostgreSQL, CrateDB, Derby, SQL Server, Db2, OceanBase and Spanner. The Derby dialect usually used for testing purpose. The field data type mappings from relational databases data types to Flink SQL data types are listed in the following table, the mapping table can help define JDBC table in Flink easily.
```
davidradl, line 711: nits:
 connect -> the connection
which uses dialect like -> using dialects e.g.
The Derby dialect usually used for testing purpose. -> The Derby dialect is usually used for testing purpose.
 JDBC table -> a  JDBC table

docs/content/docs/connectors/table/jdbc.md
laughingman7743: I have written the same URL as the official document.
- https://cloud.google.com/spanner/docs/jdbc-drivers

The Jar files can be downloaded from the following.
- https://repo1.maven.org/maven2/com/google/cloud/google-cloud-spanner-jdbc/2.27.1/
- https://repo1.maven.org/maven2/com/google/cloud/google-cloud-spanner-jdbc/2.27.1/google-cloud-spanner-jdbc-2.27.1-single-jar-with-dependencies.jar
- https://repo1.maven.org/maven2/com/google/cloud/google-cloud-spanner-jdbc/2.27.1/google-cloud-spanner-jdbc-2.27.1.jar

docs/content/docs/connectors/table/jdbc.md
laughingman7743: Fixed: https://github.com/apache/flink-connector-jdbc/pull/156/commits/3c7d2258ab0fe8bf56a60722de09b6c7e00c5350

laughingman7743:
Does this repository not have an owner? Many PRs adding dialects are being left unreviewed. Does this mean that contributions are not accepted?

laughingman7743:
## CI Failure Investigation

I investigated the CI failure in the `compile_and_test (2.1.1, 17)` job and found the following:

### Error Details

**Failed Test:** `JdbcSourceITCase.testReadWithoutParallelismWithParamsProvider`

**Root Cause Exception:**
```
java.sql.SQLNonTransientConnectionException: No current connection.
  at org.apache.flink.connector.jdbc.core.datastream.source.reader.JdbcSourceSplitReader.openResultSetForSplitWhenAtLeastOnce
  at org.apache.flink.connector.jdbc.core.datastream.source.reader.JdbcSourceSplitReader.openResultSetForSplitWhenAtMostOnce
  at org.apache.flink.connector.jdbc.core.datastream.source.reader.JdbcSourceSplitReader.openResultSetForSplit
```

### Analysis

1. **Not Spanner-specific**: This failure occurs in the `flink-connector-jdbc-core` module's test, not in Spanner-specific code.

2. **Derby JDBC Connection Issue**: The Derby JDBC connection is being closed unexpectedly during ResultSet processing.

3. **Branch Status**: This branch is already up-to-date with the latest main branch (commit 669314d), including recent hotfixes:
   - 669314d: Apply the correct version matrix in weekly CI
   - 45cbf78: Change Flink matrix to resolve Flink correctly (#179)

4. **Recent main branch CI**: The main branch CI has been passing after these hotfixes.

### Local Test Results

I ran the same test locally with both Flink versions:

**Flink 2.1.1:**
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.255 s
[INFO] BUILD SUCCESS
```

**Flink 2.0.1:**
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.325 s
[INFO] BUILD SUCCESS
```

Both tests passed successfully in the local environment.

### Conclusion

This appears to be a **CI environment-specific issue** rather than a code problem:
- The test passes locally with both Flink 2.1.1 and 2.0.1
- The issue is in `flink-connector-jdbc-core`, not Spanner-specific code
- May be related to timing, resource constraints, or containerization in the CI environment

### Recommendation

- Re-run the CI to see if it's a transient issue
- This failure should not block the Spanner dialect PR as it's unrelated to the Spanner implementation

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/statement/FieldNamedPreparedStatementImpl.java
```diff
@@ -170,6 +174,18 @@ public void setObject(int fieldIndex, Object x) throws SQLException {
...
+    public void setArray(int fieldIndex, Array x) throws SQLException {
+        for (int index : indexMapping[fieldIndex]) {
+            statement.setArray(index, x);
+        }
+    }
+
+    @Override
+    public Array createArrayOf(String typeName, Object[] elements) throws SQLException {
+        return connection.createArrayOf(typeName, elements);
+    }
```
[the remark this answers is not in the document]
laughingman7743: This implementation introduces the first full ARRAY type write support in the
flink-connector-jdbc project. Currently, PostgreSQL dialect only supports reading
ARRAY types (throws `IllegalStateException` on write - see
[PostgresDialectConverter.java:61-67](https://github.com/apache/flink-connector-jdbc/blob/v4.0.0/flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/dialect/PostgresDialectConverter.java#L61-L67)).

If this change is merged, the implementation pattern here could serve as a reference
for adding ARRAY write support to PostgreSQL and other dialects in the future.

laughingman7743:
Rebased onto the latest main to resolve the stale status. This PR is still active and awaiting review. Any feedback would be greatly appreciated.

RocMarshal:
Hi, @laughingman7743 Could you help Run 'mvn spotless:apply' before the next review?
Thanks

laughingman7743:
Hi @RocMarshal, thanks for taking a look!

I've run `mvn spotless:apply` and pushed the fix (the violation was a comment line in `SpannerCatalog.java` exceeding the line-length limit). I also rebased the branch onto the latest main. `spotless:check` now passes for all modules locally.

It looks like the CI needs maintainer approval to run — could you kindly approve the workflow when you have a chance? Thanks again for the review! 🙏

laughingman7743:
### CI failure is a pre-existing flaky test on `main`, not caused by this PR

The red job (https://github.com/apache/flink-connector-jdbc/actions/runs/27995114250/job/82864660173) fails in `DerbyDynamicTableSourceITCase.testLimit`:

```
Caused by: java.sql.SQLNonTransientConnectionException: No current connection.  (ERROR 08003)
  at ...EmbedPreparedStatement.executeQuery(...)
```

**This is not introduced by this PR:**
- This branch is rebased on `af994651`. CI for that exact commit on `main` (https://github.com/apache/flink-connector-jdbc/actions/runs/27738188350) fails in the **same `DerbyDynamicTableSourceITCase` class with the same `ERROR 08003: No current connection`**, only a different method (`testProject`).
- The latest `main` (https://github.com/apache/flink-connector-jdbc/actions/runs/27966883251) also fails CI on another flaky core ITCase (`JdbcSourceStreamRelatedITCase.testAtLeastOnceWithFailure`, 30s `TimeoutException`).
- None of the failing test files are touched by this PR; changes are isolated to the new `flink-connector-jdbc-spanner` module plus catalog/dialect/lineage in core.

### Root cause of the flaky `testLimit`

`testLimit` runs `SELECT * FROM t LIMIT 1` over a source split into **2 partitions** (`scan.partition.num=2`). With `LIMIT 1` the job completes after the first row and cancels the source while the split-fetcher thread is still opening the **second** split. Cancellation tears the JDBC connection down, so the in-flight `JdbcSourceSplitReader.openResultSetForSplitWhenAtLeastOnce()` -> `statement.executeQuery()` runs against an already-closed connection -> `08003: No current connection`.

Two gaps in `JdbcSourceSplitReader` make this fatal and racy:
- `wakeUp()` is a no-op (https://github.com/apache/flink-connector-jdbc/blob/main/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReader.java), so there is no cooperative cancellation — the reader keeps opening the next split during shutdown.
- `fetch()` rethrows any `SQLException` as a fatal `RuntimeException`, so a benign shutdown-time "connection closed" error fails the whole job.

### How the fix differs from the closed PR https://github.com/apache/flink-connector-jdbc/pull/191

That PR guarded only `resultSet.isClosed()` before `extract()` / `resultSet.next()` inside the record loop. But per the stack trace the failure is at `executeQuery()` when **opening the next split**, which that PR does not touch — so it would not reliably fix this. It was closed without merge.

The fix I'd propose is different and targets the actual cause — cooperative cancellation:
1. `wakeUp()` / `close()` set a `volatile boolean wakeup` flag.
2. `checkSplitOrStartNext()` checks the flag before opening a new split (the `executeQuery()` path) and returns "no more records" instead of starting a query during shutdown.
3. In `fetch()`, if a `SQLException` occurs while the reader is shutting down (flag set / thread interrupted), treat it as a graceful end-of-split instead of rethrowing — this closes the residual race window where cancellation lands during the Derby call.

This covers the `executeQuery()` failure point that #191 missed, and removes the "shutdown error -> job failure" behavior. I'm happy to open a separate, focused PR + JIRA for this so both this PR and `main` go green.

laughingman7743:
Follow-up on the root cause above: I filed [FLINK-39975](https://issues.apache.org/jira/browse/FLINK-39975) and opened https://github.com/apache/flink-connector-jdbc/pull/200 to fix the flaky `DerbyDynamicTableSourceITCase.testLimit`.

The fix makes `JdbcSourceSplitReader` recover from a connection that is closed while opening a split during source cancellation: it re-establishes the connection and retries the split-open (bounded, mirroring the retry/reconnect handling in `JdbcOutputFormat.flush()`). It is non-masking — a genuine query error on a healthy connection is rethrown immediately, and an unreachable database still fails the job after the retry budget. A deterministic regression test reproduces the exact `08003 "No current connection"` error (the test fails without the fix and passes with it).

Once #200 lands on `main`, this flaky test should stop reddening unrelated PRs (including this one).

RocMarshal:
Hi, @davidradl @eskabetxe 
Could you help take a look if you had the free time ? thx

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialectConverter.java
```diff
@@ -0,0 +1,223 @@
...
+                            : TimestampData.fromTimestamp((Timestamp) val);
+        } else {
+            return super.createInternalConverter(type);
+        }
+    }
+
+    @Override
+    protected JdbcSerializationConverter createNullableExternalConverter(LogicalType type) {
+        LogicalTypeRoot root = type.getTypeRoot();
+        if (root == LogicalTypeRoot.ARRAY) {
```
MartijnVisser: Overriding `createNullableExternalConverter` drops the null check, so a NULL array fails with an NPE on `arrayData.size()`. Overriding `createExternalConverter` keeps the parent wrapper.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialectConverter.java
```diff
@@ -0,0 +1,223 @@
...
+        } else if (root == LogicalTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE) {
+            // Spanner stores TIMESTAMP as an absolute instant; write TIMESTAMP_LTZ like a plain
+            // timestamp. JdbcTypeUtil has no mapping for TIMESTAMP_LTZ, so the null branch is
+            // handled here with Types.TIMESTAMP instead of the parent's nullable wrapper.
+            final int precision = LogicalTypeChecks.getPrecision(type);
+            return (val, index, statement) -> {
+                if (val == null || val.isNullAt(index)) {
+                    statement.setNull(index, Types.TIMESTAMP);
+                } else {
+                    statement.setTimestamp(index, val.getTimestamp(index, precision).toTimestamp());
```
MartijnVisser: With `-Duser.timezone=Asia/Tokyo`, `TO_TIMESTAMP_LTZ(0, 3)` is stored as 1969-12-31T15:00:00Z; reading shifts it back. I think this wants `toInstant()` here and `fromInstant` on read.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialect.java
```diff
@@ -0,0 +1,134 @@
...
+        //  PrimitiveArrayTypeInfo.BYTE_PRIMITIVE_ARRAY_TYPE_INFO in
+        // LegacyTypeInfoDataTypeConverter.
+
+        return EnumSet.of(
+                LogicalTypeRoot.CHAR,
+                LogicalTypeRoot.VARCHAR,
+                LogicalTypeRoot.BOOLEAN,
+                LogicalTypeRoot.VARBINARY,
+                LogicalTypeRoot.DECIMAL,
+                LogicalTypeRoot.TINYINT,
```
MartijnVisser: Spanner only returns INT64, so an INT column fails on read with a ClassCastException. Narrow in `createInternalConverter` or drop the smaller integer types here.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/statement/FieldNamedPreparedStatement.java
```diff
@@ -254,6 +255,21 @@ static FieldNamedPreparedStatement prepareStatement(
      */
     void setObject(int fieldIndex, Object x) throws SQLException;
 
+    /**
+     * Sets the designated parameter to the given <code>java.sql.Array</code> object. The driver
+     * converts this to an SQL <code>ARRAY</code> value when it sends it to the database.
+     *
+     * @see PreparedStatement#setArray(int, Array)
+     */
+    void setArray(int fieldIndex, Array x) throws SQLException;
```
MartijnVisser: This interface is @PublicEvolving, so new abstract methods break external implementors. Could these be default methods, with `createArrayOf` going through `statement.getConnection()`?

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
```diff
@@ -130,7 +130,30 @@ public AbstractJdbcCatalog(
...
+     *
+     * @param skipUrlValidation unused, exists only to differentiate the constructor signature
+     */
+    protected AbstractJdbcCatalog(
+            ClassLoader userClassLoader,
+            String catalogName,
+            String defaultDatabase,
+            String baseUrl,
+            Properties connectionProperties,
+            boolean skipUrlValidation) {
```
MartijnVisser: An ignored flag that only changes the signature is hard to keep on a @PublicEvolving class. I think a protected hook resolving the default database reads better.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/catalog/SpannerCatalog.java
```diff
@@ -0,0 +1,345 @@
...
+        // Skip the parent's USER_KEY/PASSWORD_KEY validation.
+    }
+
+    private ConnectionOptions getConnectionOptions() {
+        return ConnectionOptions.newBuilder().setUri(defaultUrl.replace("jdbc:", "")).build();
+    }
+
+    private SpannerOptions getSpannerOptions(ConnectionOptions options) {
+        SpannerOptions.Builder builder =
+                SpannerOptions.newBuilder().setProjectId(options.getProjectId());
```
MartijnVisser: This drops the base-url credentials, so `listDatabases` (and thus `databaseExists`, `tableExists`) uses default credentials. Passing `options.getCredentials()` should keep them aligned.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/lineage/SpannerLocationExtractorFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import io.openlineage.client.utils.jdbc.OverridingJdbcExtractor;
+
+/** Implementation of {@link JdbcLocationExtractorFactory} for Spanner. */
+@Internal
+public class SpannerLocationExtractorFactory implements JdbcLocationExtractorFactory {
+
+    @Override
+    public JdbcExtractor createExtractor() {
+        // TODO: Replace with SpannerJdbcExtractor when OpenLineage adds native Spanner support
+        return new OverridingJdbcExtractor("cloudspanner", "9010");
```
MartijnVisser: 9010 is the emulator port, and every `spanner.googleapis.com` database gets the namespace `cloudspanner://spanner.googleapis.com:9010`. Could this include project, instance and database, with a test?

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -481,7 +486,8 @@ The JDBC catalog supports the following options:
 - `base-url`: required (should not contain the database name)
   - for Postgres Catalog this should be `"jdbc:postgresql://<ip>:<port>"`
   - for MySQL Catalog this should be `"jdbc:mysql://<ip>:<port>"`
-  - for OceanBase Catalog this should be `jdbc:oceanbase://<ip>:<port>`
+  - for OceanBase Catalog this should be `"jdbc:oceanbase://<ip>:<port>"`
+  - for Spanner Catalog this should be `"jdbc:cloudspanner:/projects/<project>/instances/<instance>/databases/<database>"`
```
MartijnVisser: The catalog appends the database here, giving `.../databases/<database>/<db>`; the tests end base-url in `/databases/`. Same in zh.

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -887,6 +902,7 @@ Flink supports connect to several databases which uses dialect like MySQL, Oracl
         <code>TINYINT(1)</code></td>
       <td></td>
       <td><code>BOOLEAN</code></td>
+      <td><code>BOOLEAN</code></td>
```
MartijnVisser: Spanner calls this `BOOL`, and the cell lands after the Flink SQL column. Same in the DATE row.

flink-connector-jdbc-spanner/pom.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
MartijnVisser: Needs the ASF header, as does `SpannerTableRow.java` (`apache-rat:check`). On main the testcontainers artifacts are `testcontainers-gcloud` and `testcontainers-jdbc`.

MartijnVisser (changes requested):
@laughingman7743 Thanks. CI never reached this module because core failed first, so this needs a rebase and a green run. Please use `[FLINK-37288][jdbc]` on every commit (the lineage commit carries FLINK-34467), squash the fixups, and put the core changes in their own commits.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialectConverter.java
laughingman7743: Good catch. Arrays are now handled in `createExternalConverter`, so the parent's nullable wrapper takes care of NULL arrays. Added a row with NULL arrays to `SpannerDynamicTableSinkITCase#testArray`.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialectConverter.java
laughingman7743: Thanks, reproduced exactly as described (stored as `1969-12-31T15:00:00Z` under `Asia/Tokyo`). TIMESTAMP_LTZ is now written with `Timestamp.from(toInstant())` and read with `TimestampData.fromInstant`, including array elements. Added `testTimestampLtzWithNonUtcDefaultTimeZone`, which fails without the fix.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/dialect/SpannerDialect.java
laughingman7743: Went with narrowing in `createInternalConverter`: TINYINT/SMALLINT/INTEGER values are narrowed from the INT64 `Long`, and out-of-range values fail with an `ArithmeticException` instead of being silently truncated. The source ITCase now reads INT/SMALLINT/TINYINT columns with their min/max values.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/statement/FieldNamedPreparedStatement.java
laughingman7743: Both are now default methods throwing `SQLFeatureNotSupportedException`, so existing implementors are unaffected. `FieldNamedPreparedStatementImpl` implements them, and `createArrayOf` goes through `statement.getConnection()`; the extra connection field is gone.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
laughingman7743: Replaced the flag with a protected constructor that takes a `BiFunction<String, String, String>` default database resolver. Since `AbstractCatalog` validates the default database in its constructor, the resolution has to happen before `super(...)`, so an overridable instance method was not an option. The public constructors pass `AbstractJdbcCatalog::validateJdbcUrl`, and `SpannerCatalog` passes its own static resolver.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/catalog/SpannerCatalog.java
laughingman7743: `getSpannerOptions` now sets `options.getCredentials()`, so the admin client uses the same credentials as the JDBC connections (the application default credentials when none are configured, `NoCredentials` for the emulator). Added unit tests for `;credentials=<key file>` and `;oauthToken=...`.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/lineage/SpannerLocationExtractorFactory.java
laughingman7743: Replaced `OverridingJdbcExtractor` with a `SpannerJdbcExtractor`. It follows the OpenLineage naming convention `spanner://{projectId}:{instanceId}` and appends the database, e.g. `spanner://my-project:my-instance/my-db`, since the connector only uses the table name as the dataset name. Added `SpannerJdbcExtractorTest` covering the default host, `spanner.googleapis.com`, the emulator, and a base-url without a database.

docs/content/docs/connectors/table/jdbc.md
laughingman7743: Fixed in both en and zh: the base-url now ends with `/databases/`.

docs/content/docs/connectors/table/jdbc.md
laughingman7743: Fixed in both en and zh. While checking the table, I also found that the TIMESTAMP WITH TIMEZONE row was missing its Spanner cell, so I added it. I also removed the incorrect `NUMERIC` entry for BIGINT UNSIGNED and changed `TIMESTAMP [(p)]` to `TIMESTAMP`. Every row now has 11 cells.

flink-connector-jdbc-spanner/pom.xml
laughingman7743: Added the ASF headers to `pom.xml` and `SpannerTableRow.java`, and switched to `testcontainers-jdbc`. The gcloud module turned out to be unused (the tests use their own `SpannerEmulatorJdbcContainer`), so I dropped it.

laughingman7743:
@MartijnVisser Thanks a lot for the thorough review! I've rebased onto the latest main and addressed all comments. As requested, the history is now 5 commits, all prefixed with `[FLINK-37288][jdbc]`. The 3 core changes each have their own commit, followed by the Spanner module and the docs.

Besides the review items, I bumped `google-cloud-spanner-jdbc` from 2.26.1 to 2.44.0 and the emulator image to 1.5.58. Two things were needed for the new driver:
- Newer drivers return parameterized type names such as `ARRAY<BOOL>` from `getColumnTypeName`, so the catalog type mapper now strips the type parameters.
- Since 2.32.0, all connections share a single multiplexed session. The emulator only supports one read-write transaction at a time, so `testUpsert` failed with parallel sink subtasks on the emulator. The Spanner ITCase now runs that test with parallelism 1. This limitation is specific to the emulator; real Spanner supports concurrent transactions on multiplexed sessions.

Locally, the core, Spanner and architecture tests pass on JDK 17. Could you please approve the CI workflow when you have a chance?

laughingman7743:
@MartijnVisser Thanks for approving the CI run. The Flink 2.3.0 job failed in `SpannerDynamicTableSinkITCase`, which I've fixed and force-pushed. The history is still the same 5 commits.

**Cause:** Since Flink 2.3.0 (FLIP-558), writing insert-only data into a sink table with a primary key requires an `ON CONFLICT` clause. Spanner tables always have a primary key, so `testAppend`, `testReal`, `testArray` and `testTimestampLtzWithNonUtcDefaultTimeZone` hit the new validation. FLINK-40725 only adjusted the tests of other dialects, and I had only run the tests locally against the default Flink 2.1.3, so I missed this.

**Fix (tests only):**
- `JdbcDynamicTableSinkITCase` (core): made `isFlip558Enabled()` protected and added an `appendOnConflictClause()` hook that `testAppend` appends to its INSERT statement. It returns an empty string by default, so dialects whose append output table has no primary key are unaffected. Their `testAppend` already passed on Flink 2.3.0 in this CI run.
- `SpannerDynamicTableSinkITCase`: overrides the hook to return ` ON CONFLICT DO DEDUPLICATE` on Flink >= 2.3, the same clause as FLINK-40725, and uses it in the Spanner-specific sink tests.

The core, Spanner and architecture tests now pass locally against both Flink 2.1.3 and 2.3.0.

The other failures in that run were not caused by this PR. The Flink 2.1.3 and 2.2.1 jobs were cancelled by fail-fast after the 2.3.0 job failed, and `backwards_compatibility_test (2.0-SNAPSHOT)` failed because Maven Central returned a transient `500 Internal Server Error` while resolving `maven-compiler-plugin`.

Could you please approve the workflow run again? Thanks!

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/catalog/SpannerTypeMapper.java
```diff
@@ -0,0 +1,170 @@
...
+                return DataTypes.DECIMAL(38, 9);
+            case SPANNER_STRING:
+            // The default column display size is returned for precision,
+            // so it is handled as a STRING type, not a VARCHAR type.
+            // https://github.com/googleapis/java-spanner-jdbc/blob/v2.26.1/src/main/java/com/google/cloud/spanner/jdbc/JdbcResultSetMetaData.java#L157
+            case SPANNER_JSON:
+            case SPANNER_STRUCT:
+                return DataTypes.STRING();
+            case SPANNER_TIMESTAMP:
+                return DataTypes.TIMESTAMP();
```
MartijnVisser: Spanner TIMESTAMP is an instant with nanoseconds. Through the catalog, `1970-01-01T00:00:00.123456789Z` reads as `1970-01-01 09:00:00.123456` with the JVM in Asia/Tokyo. I think it should be `TIMESTAMP_LTZ(9)`, as Postgres does for `timestamptz`, arrays included.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
```diff
@@ -141,6 +167,10 @@ public AbstractJdbcCatalog(
         this.connectionProperties = Preconditions.checkNotNull(connectionProperties);
         this.defaultUrl = this.urlFunction.apply(defaultDatabase);
 
+        validateConnectionProperties(connectionProperties);
+    }
+
+    protected void validateConnectionProperties(Properties connectionProperties) {
```
MartijnVisser: This and `calculateUrlFunction` are now protected API on a `@PublicEvolving` class and run from the constructor, before subclass fields are set. That's worth a Javadoc.

MartijnVisser (changes requested):
@laughingman7743 I've checked the earlier comments on b538319: the new NULL array, TIMESTAMP_LTZ and INT tests fail on the old converter and pass now, and CI is green. I'd like to settle the TIMESTAMP mapping before merging.

docs/content/docs/connectors/table/jdbc.md
MartijnVisser: The table still points at the plain `google-cloud-spanner-jdbc` jar, which doesn't contain `SpannerOptions` or `ConnectionOptions`. Can it name the `single-jar-with-dependencies` classifier? Same in zh.

laughingman7743: Good point, the catalog needs `SpannerOptions` and `ConnectionOptions`, which are only bundled in the shaded jar. The driver table now names `google-cloud-spanner-jdbc` with the `single-jar-with-dependencies` classifier and links to the Maven Central directory, in both en and zh.

flink-connector-jdbc-spanner/src/main/java/org/apache/flink/connector/jdbc/spanner/database/catalog/SpannerTypeMapper.java
laughingman7743: Agreed. The catalog now maps Spanner `TIMESTAMP` to `TIMESTAMP_LTZ(9)` and `ARRAY<TIMESTAMP>` to `ARRAY<TIMESTAMP_LTZ(9)>`, like Postgres `timestamptz`. The read path already converts through `Instant` since the previous fix. `SpannerCatalogITCase` now uses nanosecond test data and expects `Instant` values. I also added `testTimestampIsIndependentOfDefaultTimeZone`, which reads through the catalog with the JVM in `Asia/Tokyo`. With the old mapping, the catalog tests failed with `2025-01-01T09:00:00.123456789`. The docs type table now lists Spanner `TIMESTAMP` in the `TIMESTAMP WITH TIMEZONE` row.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
laughingman7743: Added Javadoc to both `validateConnectionProperties` and `calculateUrlFunction`. It describes what each hook is for and notes that it is called from the constructor before subclass fields are initialized, so implementations must only depend on the given argument.

PR #161, opened 2025-04-10
[FLINK-33463][Connector/JDBC] Support the implementation of dynamic source tables based on the new source

RocMarshal:


RocMarshal:
Hi, @leonardBang @snuyanzin Could you help take a look ? Thank you~

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/table/RowDataResultExtractor.java
ferenc-csaky: I'd rather put this class into the `extractor` package, next to its interface and `RowExtractor`. Based on the current grouping IMO it makes more sense.

ferenc-csaky (changes requested):
Thanks, mostly looks good, added 2 suggestions.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/table/RowDataResultExtractor.java
RocMarshal: @ferenc-csaky What about put `RowDataResultExtractor` into pacakge `org.apache.flink.connector.jdbc.core.table.source.extractor` ?

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/extractor/ResultExtractor.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.flink.connector.jdbc.core.datastream.source.reader.extractor;
```
eskabetxe: This is a public interface we should not move it so easy.
If we are moving it we need to keep an class (deprecated) in origin that extends the moved one to keep retro compatibility, but I would avoid database as is the API for connectors (the things that connectors need to implement)

I understand that this refactor is because of the new implementation, but being table using the datastream why don't we put the new clase on the current package? that is a valid implementation also for datastream and useful in some cases (reading from jdbc and sending to paimon/iceberg as example)

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/extractor/RowDataResultExtractor.java
eskabetxe: org.apache.flink.connector.jdbc.core.database is basically the API for database connectors..

so I think we should avoid that package

eskabetxe: > I'd rather put this class into the `extractor` package, next to its interface and `RowExtractor`. Based on the current grouping IMO it makes more sense.

I would say that this makes more sense to me, as tableApi is using datastreamAPI and ResultExtractor is created for datastream use case, and I can see this new implementation used on datastream use cases also (reading from jdbc and sending to paimon/iceberg as example)

RocMarshal:
Thanks @eskabetxe @ferenc-csaky .
Whether we move both the interface and its implementations into the database package while ensuring compatibility, or```I'd rather put this class into the extractor package, next to its interface and RowExtractor. Based on the current grouping IMO it makes more sense.```,
I think either option is fine. However, I’m still curious—which approach will we ultimately choose? From the previous discussions, I didn’t see a clear conclusion.

In other words, which package path are we currently leaning toward for placing ResultExtractor.java and its implementation classes?
Thanks for your confirmation in advance.

PR #169, opened 2025-05-25
FLINK-36236 Fix overriding the connection provider in JdbcSourceBuilder

rajucomp:
#### Summary

This PR fixes an issue where a custom `JdbcConnectionProvider` passed via `JdbcSourceBuilder#setConnectionProvider` was unintentionally overwritten during the `build()` process. This prevented users from injecting their own connection management logic (e.g., for testing or advanced connection pooling scenarios).

#### Changes

- **`JdbcSourceBuilder`**
  - Modified `build()` method to only set a default `SimpleJdbcConnectionProvider` **if** no custom provider has been set.
  
- **`JdbcSource`**
  - Exposed a `getConnectionProvider()` method with `@VisibleForTesting` to enable access from test code.

- **Test Coverage**
  - Added a unit test `testSetConnectionProvider()` to:
    - Verify that `null` values are rejected.
    - Assert that the `JdbcSource` retains and uses the explicitly provided `JdbcConnectionProvider`.

#### Motivation

This change ensures the builder pattern behaves predictably: user-provided values should not be silently overridden. This is especially important when dependency injection or custom behavior is expected during source initialization.

### DIFF
--- a/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSource.java
+++ b/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSource.java
@@ -162,18 +162,23 @@ public class JdbcSource<OUT>
     }
 
     @VisibleForTesting
-    public TypeInformation<OUT> getTypeInformation() {
+    TypeInformation<OUT> getTypeInformation() {
         return typeInformation;
     }
 
     @VisibleForTesting
-    public Configuration getConfiguration() {
+    Configuration getConfiguration() {
         return configuration;
     }
 
     @VisibleForTesting
-    public ResultExtractor<OUT> getResultExtractor() {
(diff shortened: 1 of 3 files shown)


rajucomp:
@eskabetxe Can I get a review please ? Thanks.

eskabetxe (approved):
LGTM. 

My only concern is the change in visibility of some methods in a @PublicEvolving class. I understand they're annotated with @VisibleForTesting, but we should be cautious with such changes.

rajucomp:
@@eskabetxe  the checks have passed. Could you merge the PR ? I don't think I have the permission to do so. Thanks!

ferenc-csaky (approved):
LGTM! Based on the definition (pasted below) `@PublicEvolving` I see no harm in restricting the visibility of those getters.
```
Classes and methods with this annotation are intended for public use and have stable behavior. However, their interfaces and signatures are not considered to be stable and might be changed across versions. 
```

@rajucomp this supersedes #150, right?

PR #142, opened 2024-09-17
fix PostgreSQL minimum timestamp precision

kgeis:
fixes [FLINK-36303](https://issues.apache.org/jira/browse/FLINK-36303)

PR #172, opened 2025-06-05
[FLINK-36303] PostgreSQL timestamp precision 0 not supported

kgeis:
This is a followup to #142 to make the integration test match the original code change.

### DIFF
--- a/flink-connector-jdbc-cratedb/src/test/java/org/apache/flink/connector/jdbc/cratedb/database/dialect/CrateDBDialectTest.java
+++ b/flink-connector-jdbc-cratedb/src/test/java/org/apache/flink/connector/jdbc/cratedb/database/dialect/CrateDBDialectTest.java
@@ -51,7 +51,7 @@ class CrateDBDialectTest extends JdbcDialectTest implements CrateDBTestBase {
                 createTestItem("BINARY", "The CrateDB dialect doesn't support type: BINARY(1)."),
                 createTestItem(
                         "TIMESTAMP(9) WITHOUT TIME ZONE",
-                        "The precision of field 'f0' is out of the TIMESTAMP precision range [1, 6] supported by CrateDB dialect."),
+                        "The precision of field 'f0' is out of the TIMESTAMP precision range [0, 6] supported by CrateDB dialect."),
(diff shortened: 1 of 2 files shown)


1996fanrui (approved):
@kgeis  Thanks for the quick fix!

LGTM assuming the CI is green.

1996fanrui (changes requested):
It's still failed.

https://github.com/apache/flink-connector-jdbc/actions/runs/15474420512/job/43567756737?pr=172#step:16:1118


```
Expecting throwable message:
  "The precision of field 'f0' is out of the TIMESTAMP precision range [0, 6] supported by CrateDB dialect."
to contain:
  "The precision of field 'f0' is out of the TIMESTAMP precision range [1, 6] supported by CrateDB dialect."
```

kgeis:
Sorry, I didn't realize that change would have impacts elsewhere.

Reviewing the code, I question the decision to have the CrateDB dialect extend the PostgreSQL. Just because they talk the same wire protocol doesn't mean their SQL or data types will be the same. For example, `CREATE TABLE T (f0 TIMESTAMP(3))` works on PostgreSQL 17 but CrateDB 5.10 complains
```
Error!

SQLParseException[The 'timestamp' type doesn't support type parameters.]
```

That said, I was able to get the tests running locally (at least through the CrateDB tests), and when I updated the CrateDBDialectTest to follow the PostgreSQLDialectTest, they passed.

1996fanrui (commented):
Hi @matriv  @eskabetxe , would you mind double checking this PR as well? Thanks in advance.

The background could be found from https://github.com/apache/flink-connector-jdbc/pull/142 and FLINK-36303

eskabetxe (approved):
I'm ok with the change as we are fixing on postgres and not introducing nothing new on CrateDB

We could check after (maybe in another Jira) the CrateDB extension, for timestamp we should fix this, as its only allowed timestmap, the precision should be 3 on min and max.

1996fanrui (commented):
Thanks @kgeis for the update, and @eskabetxe for the review, merging

kgeis:
Please release this someday.

PR #176, opened 2025-08-29
Add Postgresql UUID catalog support

kcolford:


flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/database/catalog/PostgresTypeMapper.java
```diff
@@ -154,6 +155,8 @@ protected DataType getMapping(String pgType, int precision, int scale) {
                 return DataTypes.ARRAY(DataTypes.CHAR(precision));
             case PG_CHARACTER_VARYING:
                 return DataTypes.VARCHAR(precision);
+            case PG_UUID:
+                return DataTypes.VARCHAR(36);
```
davidradl, line 159: could we add tests to drive this please for Postgres?

PR #198, opened 2026-06-21
[FLINK-38046][Connector/JDBC] Solve the issue of missing one piece of data

och5351:
# What is the purpose of the change

When using partitioned scan in Flink JDBC, if scan.partition.lower-bound and scan.partition.upper-bound are very large values (near Long range), integer arithmetic overflow causes the last partition to be silently dropped, resulting in one batch of data being lost.

This fix was originally identified and implemented by @wangxiaojing in #174. Since the PR had been inactive for an extended period with no release progress, I am resubmitting it to move things forward. All credit for the investigation and fix goes to @wangxiaojing.

# Brief change log

- JdbcNumericBetweenParametersProvider.getParameterValues(): changed the loop to iterate batchNum - 1 times and explicitly set the last partition's upper bound to maxVal
- Added regression test testBatchMaxMinTooLarge covering the overflow scenario with large Long values

# Verifying this change

- Run NumericBetweenParametersProviderTest#testBatchMaxMinTooLarge which covers the exact overflow case (lower=2260418954055131340, upper=3875220057236942850, batchNum=3)
- All 7 existing tests in NumericBetweenParametersProviderTest pass

### DIFF
--- a/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/split/JdbcNumericBetweenParametersProvider.java
+++ b/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/split/JdbcNumericBetweenParametersProvider.java
@@ -112,11 +112,12 @@ public class JdbcNumericBetweenParametersProvider implements JdbcParameterValues
 
         Serializable[][] parameters = new Serializable[batchNum][2];
         long start = minVal;
-        for (int i = 0; i < batchNum; i++) {
+        for (int i = 0; i < batchNum - 1; i++) {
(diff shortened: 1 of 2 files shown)


och5351:
Hi, @wangxiaojing , @RocMarshal !

I've addressed the review comment regarding the test case — switched to ofBatchNum and corrected the expected values as suggested.

> https://github.com/apache/flink-connector-jdbc/pull/174#discussion_r2596822822

All tests are passing now. Could you please take another look when you have a moment?

Thank you!

RocMarshal:
LGTM +1.
Thanks @och5351 @wangxiaojing

PR #200, opened 2026-06-23
[FLINK-39975][jdbc] Recover JdbcSourceSplitReader from a connection closed during cancellation

laughingman7743:

Fix the flaky `DerbyDynamicTableSourceITCase.testLimit` (and occasionally `testProject`). When a bounded query (e.g. `LIMIT`) finishes the job early, the source is cancelled while the split fetcher is still opening the next split; the
JDBC connection validated by the provider is torn down before the statement is prepared/executed, so Derby raises `08003: No current connection` and the fatal `SQLException` fails the whole job (the test runs under
`NoRestartBackoffTimeStrategy`). See [FLINK-39975](https://issues.apache.org/jira/browse/FLINK-39975) for the full analysis.

## Brief change log

- `JdbcSourceSplitReader` now opens a split via `openResultSetForSplitWithReconnect`, a bounded retry (`MAX_CONNECTION_RETRIES = 3`).
- It retries only when the connection is actually closed (`connection.isClosed()`); before retrying it drops the dead connection and the stale statement/result set and re-establishes via the provider. A query error on a healthy
connection is rethrown immediately, and an unreachable database still fails after the retry budget — real errors are not masked.
- `wakeUp()` / `fetch()` are unchanged.

## Verifying this change

- Added `JdbcSourceSplitReaderTest#testFetchReconnectsWhenConnectionClosedWhileOpeningSplit`, a deterministic regression test that injects an already-closed connection, reproducing the exact `08003` error (it fails without this
change) and asserting the reader re-establishes the connection and reads the whole split.
- Added `testFetchRethrowsImmediatelyWhenConnectionStaysOpen` (a query error on a healthy connection is rethrown without reconnect) and `testFetchFailsAfterExhaustingReconnectRetries` (gives up after the retry budget instead of
looping forever).
- `DerbyDynamicTableSourceITCase` passes repeated runs.

- The serializers: no
- The runtime per-record code paths (performance sensitive): no — the retry only triggers on a closed-connection error while opening a split

## Documentation

### DIFF
--- a/flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
+++ b/flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
@@ -34,6 +34,8 @@ import org.apache.flink.connector.testutils.source.reader.TestingReaderContext;
 
 import org.junit.jupiter.api.Test;
 
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


RocMarshal (approved):
LGTM +1.
Thanks @laughingman7743

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
```diff
@@ -144,4 +146,120 @@ void testFetch() throws Exception {
...
+
+    @Test
+    void testFetchReconnectsWhenConnectionClosedWhileOpeningSplit() throws Exception {
+        // The provider hands back an already-closed connection the first time, so the reader's
+        // first use of it fails (mimicking the connection being torn down, e.g. on source
+        // cancellation, before the reader finishes opening the split). The reader must
+        // re-establish the connection and read the whole split.
+        CountingConnectionProvider provider = new CountingConnectionProvider(connectionProvider, 1);
+        JdbcSourceSplitReader<TestEntry> splitReader = newReader(provider, split);
+        try {
```
RocMarshal: Could the lines be 
```
try(JdbcSourceSplitReader<TestEntry> splitReader = newReader(provider, split)) {
....
}

// deleted
// xxx.close()

```

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
```diff
@@ -144,4 +146,120 @@ void testFetch() throws Exception {
...
+            RecordAndOffset<TestEntry> recordAndOffset = fetched.nextRecordFromSplit();
+            while (recordAndOffset != null) {
+                records.add(recordAndOffset.record);
+                recordAndOffset = fetched.nextRecordFromSplit();
+            }
+            assertThat(records).hasSize(TEST_DATA.length);
+            // The first (closed) connection plus exactly one re-established one: a reconnect ran.
+            assertThat(provider.establishCount).isEqualTo(2);
+        } finally {
+            splitReader.close();
```
RocMarshal: if so, deleted

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
```diff
@@ -144,4 +146,120 @@ void testFetch() throws Exception {
...
+
+    @Test
+    void testFetchRethrowsImmediatelyWhenConnectionStaysOpen() throws Exception {
+        // A query error on a healthy (open) connection must be rethrown immediately, with no
+        // reconnect attempt.
+        CountingConnectionProvider provider = new CountingConnectionProvider(connectionProvider, 0);
+        JdbcSourceSplit invalidSplit =
+                new JdbcSourceSplit("1", "select * from NON_EXISTENT_TABLE", null, null);
+        JdbcSourceSplitReader<TestEntry> splitReader = newReader(provider, invalidSplit);
+        try {
```
RocMarshal: if so, ditto

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
```diff
@@ -144,4 +146,120 @@ void testFetch() throws Exception {
...
+    }
+
+    @Test
+    void testFetchFailsAfterExhaustingReconnectRetries() throws Exception {
+        // If the connection keeps being closed, the reader gives up after the retry budget
+        // instead of looping forever.
+        CountingConnectionProvider provider =
+                new CountingConnectionProvider(connectionProvider, Integer.MAX_VALUE);
+        JdbcSourceSplitReader<TestEntry> splitReader = newReader(provider, split);
+        try {
```
RocMarshal: if so, ditto

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReaderTest.java
laughingman7743: Good catch, thanks! Switched the three new tests to try-with-resources and dropped the explicit close().

PR #205, opened 2026-07-08
[hotfix] Add missing ASF license headers in backward-compatibility module

weiqingy:

The `flink-connector-jdbc-backward-compatibility` module ships two files that are missing the Apache license header every other source file in the repository carries:

- `flink-connector-jdbc-backward-compatibility/pom.xml`
- `flink-connector-jdbc-backward-compatibility/src/test/java/org/apache/flink/connector/jdbc/backward/compatibility/DynamicTableSourceTest.java`

Both files are included in the source release tarball (observed while verifying the `flink-connector-jdbc-4.1.0` RC3 source distribution; the same state is on `main`). The module's three sibling test files already carry the header.

## Why it isn't caught by CI

The module is intentionally excluded from the parent pom's `<modules>` so it can be built standalone against multiple Flink versions. Two consequences follow:

- The repository-wide Apache RAT check is bound to the `verify` phase of the root reactor, which never includes this module.
- The dedicated `backwards_compatibility.yml` workflow runs `mvn ... test` inside the module directory, which stops before the `verify` phase where RAT runs.

So neither path exercises RAT on these two files.

## Brief change log

- Add the standard ASF license header to `pom.xml` and `DynamicTableSourceTest.java`, byte-identical to the header already present on the module's sibling files.

## Verifying this change

License-header-only change; no production or test logic is touched. Verified locally that `mvn apache-rat:check` reports both files as `AL` (approved license) after the change; before the change they were flagged as unapproved.

- The serializers: no

## Documentation

### DIFF
--- a/flink-connector-jdbc-backward-compatibility/pom.xml
+++ b/flink-connector-jdbc-backward-compatibility/pom.xml
@@ -1,4 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


RocMarshal:
Thanks @weiqingy

PR #209, opened 2026-07-09
[FLINK-40108] Adding snapshot capability with split enumerator

eskabetxe:
This pull request introduces a new connection management layer for the Flink JDBC connector, adding support for pluggable connection providers and improved connection configuration. The changes include new classes for connection options, a builder for those options, a minimal `DataSource` implementation for integration with connection pools, and interfaces for connection and statement handling. Additionally, HikariCP is added as a dependency (with SLF4J excluded), and a new async enumerator is provided for snapshot splitting.

**Connection Management and Configuration:**

* Added `ConnectionOptions` and `ConnectionOptionsBuilder` classes to encapsulate and build JDBC connection configuration, including timeouts and properties. [[1]](diffhunk://#diff-f001332fa66e39c36c0cc30834dda90a0fa41810bd7cd8bae6530ee856fede0fR1-R52) [[2]](diffhunk://#diff-b839c5096307f402ada2ccd9a7a55bcae8d79a27238b19ede6116d971769ee72R1-R91)
* Introduced `ConnectionDataSource`, a minimal `DataSource` implementation that delegates driver resolution, enabling proper classloader handling with connection pools like HikariCP.
* Added `ConnectionException` for consistent error handling around connection issues.
* Added `ConnectionProvider` interface, extending the base JDBC connection provider with methods for table/column metadata discovery and query building for snapshot splitters.

**Statement and Result Handling:**

* Introduced `ResultSetMapper` and `StatementPreparer` functional interfaces for mapping JDBC results and preparing statements, respectively. [[1]](diffhunk://#diff-eefcfd1c56b12cd6ab7d02efcdbd1ae004f7b609f6a820b95aafd8ea2c05d514R1-R33) [[2]](diffhunk://#diff-8507c8062c472a991cb94c4588edcac8fc52f3727db77c494dfb2cc586fcfb5bR1-R29)

**Async Snapshot Splitting:**

* Added `AsyncSnapshotSplitterEnumerator`, a background-threaded enumerator for computing and emitting snapshot splits asynchronously, improving scalability and responsiveness during snapshotting.

**Dependency Management:**

* Added HikariCP as a dependency in `pom.xml`, with SLF4J excluded to avoid conflicts.

### DIFF
--- a/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
+++ b/flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
@@ -1,0 +1,447 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.connector.jdbc.core.datastream.connection;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.connector.jdbc.core.datastream.source.enumerator.splitter.snapshot.domain.TableColumn;
+import org.apache.flink.connector.jdbc.core.datastream.source.enumerator.splitter.snapshot.domain.TableId;
+import org.apache.flink.util.Preconditions;
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+
+import java.sql.Connection;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.time.Duration;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Base {@link ConnectionProvider} that implements the connection lifecycle, pooling, and
+ * prepared-statement plumbing shared by every dialect-specific provider. Subclasses only need to
+ * implement the dialect-specific table discovery and bound-query building methods declared by
+ * {@link ConnectionProvider} (e.g. {@code getTables}, {@code queryMinMax}, {@code
+ * queryNextChunkMax}, {@code createQueryWithBounds}, {@code newInstance}).
+ */
+@Internal
+public abstract class AbstractConnectionProvider implements ConnectionProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractConnectionProvider.class);
+
+    private static final int DEFAULT_POOL_SIZE = 4;
+    private static final int MINIMUM_POOL_SIZE = 1;
+
+    protected final ConnectionOptions jdbcOptions;
+    private int queryTimeoutSeconds;
+    private transient Driver loadedDriver;
+    private transient Connection connection;
+    private final Map<String, PreparedStatement> statementCache;
+    private transient HikariDataSource connectionPool;
+    private final boolean poolOwner;
+
+    static {
+        // Load DriverManager first to avoid deadlock between DriverManager's
+        // static initialization block and specific driver class's static
+        // initialization block when two different driver classes are loading
+        // concurrently using Class.forName while DriverManager is uninitialized
+        // before.
+        //
+        // This could happen in JDK 8 but not above as driver loading has been
+        // moved out of DriverManager's static initialization block since JDK 9.
+        DriverManager.getDrivers();
+    }
+
+    protected AbstractConnectionProvider(ConnectionOptions jdbcOptions) {
+        this.jdbcOptions = jdbcOptions;
+        this.statementCache = new ConcurrentHashMap<>();
+        this.queryTimeoutSeconds = jdbcOptions.getConnectionQueryTimeoutSeconds();
+        this.poolOwner = true;
+        try {
+            // Establish the raw connection directly rather than via getOrEstablishConnection():
+            // that method invokes the overridable onConnectionEstablished() hook, which a
+            // subclass may depend on its own fields for — fields that aren't initialized yet
+            // while this superclass constructor is still running.
+            establishConnection();
+        } catch (Exception e) {
+            throw new ConnectionException(
+                    "Failed to establish initial connection during connection provider construction.",
+                    e);
+        }
+    }
+
+    /**
+     * Creates a pooled instance that borrows its connection from the given pool. This instance does
+     * NOT own the pool and will not shut it down on close.
+     */
+    protected AbstractConnectionProvider(ConnectionOptions jdbcOptions, HikariDataSource pool) {
+        this.jdbcOptions = jdbcOptions;
+        this.statementCache = new ConcurrentHashMap<>();
+        this.queryTimeoutSeconds = jdbcOptions.getConnectionQueryTimeoutSeconds();
+        this.connectionPool = pool;
+        this.poolOwner = false;
+        try {
+            this.connection = pool.getConnection();
+        } catch (SQLException e) {
+            throw new ConnectionException("Failed to borrow connection from pool.", e);
+        }
+    }
+
+    protected synchronized HikariDataSource getOrCreatePool() {
+        if (connectionPool == null) {
+            connectionPool = createConnectionPool();
+        }
+        return connectionPool;
+    }
+
+    private HikariDataSource createConnectionPool() {
+        HikariConfig config = new HikariConfig();
+        // Provide a DataSource directly so HikariCP doesn't try to load the driver
+        // via its own classloader. In Flink, the JDBC driver lives in the user
+        // classloader and is not visible to HikariCP's threads. Using this driver
+        // delegate guarantees connections are created with the same code path as
+        // getOrEstablishConnection().
+        config.setDataSource(new ConnectionDataSource(jdbcOptions, this::getLoadedDriver));
+        config.setMinimumIdle(MINIMUM_POOL_SIZE);
+        config.setMaximumPoolSize(maxPoolSize());
+        config.setConnectionTimeout(
+                Duration.ofSeconds(jdbcOptions.getConnectionCheckTimeoutSeconds()).toMillis());
+        config.setPoolName(poolName());
+        LOG.info("Creating HikariCP connection pool with maxPoolSize={}", maxPoolSize());
+        return new HikariDataSource(config);
+    }
+
+    /** Maximum number of pooled connections. Override to change the pool size. */
+    protected int maxPoolSize() {
+        return DEFAULT_POOL_SIZE;
+    }
+
+    /** Name used for the HikariCP pool, shown in logs/metrics. */
+    protected String poolName() {
+        return getClass().getSimpleName() + "-pool";
+    }
+
+    /**
+     * Hook invoked every time a connection is (re-)established, before it's handed back to the
+     * caller. No-op by default; dialect-specific subclasses can override to re-apply
+     * connection-scoped state (e.g. re-syncing a shared snapshot transaction id).
+     */
+    protected void onConnectionEstablished() throws SQLException {}
+
+    @Override
+    public Connection getConnection() {
+        return connection;
+    }
+
+    @Nonnull
+    @Override
+    public Properties getProperties() {
+        return jdbcOptions.getProperties();
+    }
+
+    @Override
+    public boolean isConnectionValid() throws SQLException {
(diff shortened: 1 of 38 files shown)


eskabetxe:
@RocMarshal @snuyanzin can you check please

eskabetxe:
@och5351 can you help review

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/AsyncSnapshotSplitterEnumerator.java
```diff
@@ -0,0 +1,189 @@
...
+            throw new IllegalArgumentException(
+                    "Connection provider must be an instance of "
+                            + ConnectionProvider.class.getSimpleName());
+        }
+        this.connection = (ConnectionProvider) connectionProvider;
+    }
+
+    @Override
+    public final Boundedness getBoundedness() {
+        return Boundedness.CONTINUOUS_UNBOUNDED;
```
och5351, line 79: Snapshot reads are finite, so this should return BOUNDED.
With CONTINUOUS_UNBOUNDED the job never terminates after all chunks are emitted.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
```diff
@@ -0,0 +1,447 @@
...
+        }
+        establishConnection();
+        onConnectionEstablished();
+        return connection;
+    }
+
+    /**
+     * Establishes a fresh {@link #connection}. Does not invoke {@link #onConnectionEstablished()}.
+     */
+    private Connection establishConnection() throws SQLException, ClassNotFoundException {
```
och5351: It's maybe void. isn't it?

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
```diff
@@ -0,0 +1,447 @@
...
+                // caller expectation.
+                throw new SQLException("No suitable driver found for " + connectionUrl, "08001");
+            }
+        }
+        return connection;
+    }
+
+    @Override
+    public Connection reestablishConnection() throws SQLException, ClassNotFoundException {
+        closeConnection();
```
och5351, line 249: Edge case, but worth noting: on a 10s+ network outage (e.g., RDS failover), closeConnection() times out, aborts, and nulls this.connection. 

reestablishConnection() then falls through to establishConnection() -> DriverManager.getConnection(), bypassing the HikariCP pool entirely.

For pool-borrowed instances, wouldn't connectionPool.getConnection() be more appropriate here?

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/TableSplitterEnumerator.java
```diff
@@ -0,0 +1,212 @@
...
+    private final int chunkSize;
+
+    private TableColumn tablePrimaryKey;
+    private final Set<String> lineageQueries;
+
+    private TableBounds tableMinMax;
+    private Object currentLowerBound;
+    private boolean boundsInitialized;
+
+    protected TableSplitterEnumerator(TableId tableId, Set<String> columnNames, int chunkSize) {
```
och5351, line 61: DatabaseSplitterEnumerator.prepareTableSplitters() already calls connection.getTableColumns() but strips it to Set<String> before passing to the constructor. 

Then TableSplitterEnumerator.validateTableAndColumns() calls getTableColumns() again for the same table.

How about adding a second constructor that accepts pre-loaded Set<TableColumn>?

```suggestion
// standalone usage — fetches lazily as before
protected TableSplitterEnumerator(TableId tableId, Set<String> columnNames, int chunkSize) {
    ...
    this.discoveredColumns = null;
}

// called from DatabaseSplitterEnumerator with pre-loaded columns
protected TableSplitterEnumerator(TableId tableId, Set<TableColumn> preloadedColumns, int chunkSize) {
    ...
    this.discoveredColumns = preloadedColumns;
}

private void validateTableAndColumns() {
    if (this.discoveredColumns == null) {
        this.discoveredColumns = connection.getTableColumns(tableId);
    }
    ...
}
```

This avoids the redundant metadata query when called from DatabaseSplitterEnumerator.

och5351 (changes requested):
Hi, @eskabetxe !
Thank you for your contribution.
I have a just nit.
PTAL.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/AsyncSnapshotSplitterEnumerator.java
eskabetxe: It will end because the splitter when finished will tell that no more splits available.. the splits are bounded.

it continuous_bounded to allow process big tables, a table with 1T of data will be problematic witn bounded.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
eskabetxe: yes, can be void.. changed

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
eskabetxe: good catch, fixed

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/TableSplitterEnumerator.java
eskabetxe: The overloaded constructor won't compile: Set<String> and Set<TableColumn> erase to the same (TableId, Set, int) signature.

But we are calling two times per table not partitioned, I will say the cost is marginal..

eskabetxe:
@och5351 can you recheck

flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/datastream/connection/PostgresConnectionProvider.java
```diff
@@ -0,0 +1,307 @@
...
+                Connection currentConn = getConnection();
+                assert currentConn != null;
+                if (!currentConn.getAutoCommit()) {
+                    currentConn.rollback();
+                }
+                currentConn.setAutoCommit(false);
+                currentConn.setTransactionIsolation(Connection.TRANSACTION_REPEATABLE_READ);
+                LOG.info("Setting connection with snapshot id: {}", snapshotId);
+                try (PreparedStatement statement =
+                        currentConn.prepareStatement("SET TRANSACTION SNAPSHOT ?")) {
```
MartijnVisser, line 127: Postgres takes no bind parameter here; this throws `syntax error at or near "$1"` on 16. Needs the id as a literal, and `execute()`.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
```diff
@@ -0,0 +1,451 @@
...
+        }
+        establishConnection();
+        onConnectionEstablished();
+        return connection;
+    }
+
+    /**
+     * Establishes a fresh {@link #connection}. Does not invoke {@link #onConnectionEstablished()}.
+     */
+    private void establishConnection() throws SQLException, ClassNotFoundException {
```
MartijnVisser, line 227: This replaces `connection` but leaves `statementCache` bound to the old one, so a dropped connection never recovers. Verified against Postgres 16.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/DatabaseSplitterEnumerator.java
```diff
@@ -0,0 +1,217 @@
...
+    public static DatabaseSplitterEnumeratorBuilder builder() {
+        return new DatabaseSplitterEnumeratorBuilder();
+    }
+
+    @Override
+    public void start(JdbcConnectionProvider connectionProvider) {
+        initConnection(connectionProvider);
+        this.pendingTableSplitters = new ConcurrentLinkedQueue<>();
+        this.activeTableSplitters = new CopyOnWriteArrayList<>();
+        prepareTableSplitters();
```
MartijnVisser, line 75: `start()` is on the SourceCoordinator thread and this queries per table inline, stalling split assignment and checkpoints. Move it to the background thread below.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/AsyncSnapshotSplitterEnumerator.java
```diff
@@ -0,0 +1,189 @@
...
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+        }
+        closeResources();
+    }
+
+    @Override
+    public final Serializable serializableState() {
```
MartijnVisser, line 181: Returns null and `restoreState()` ignores its argument, so a restore re-emits every split. Both `final`, and the test asserts it as the contract.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/AsyncSnapshotSplitterEnumerator.java
```diff
@@ -0,0 +1,189 @@
...
+        if (!(connectionProvider instanceof ConnectionProvider)) {
+            throw new IllegalArgumentException(
+                    "Connection provider must be an instance of "
+                            + ConnectionProvider.class.getSimpleName());
+        }
+        this.connection = (ConnectionProvider) connectionProvider;
+    }
+
+    @Override
+    public final Boundedness getBoundedness() {
```
MartijnVisser, line 78: Back to the open thread above: boundedness does not control split production, so table size is not the reason. BATCH mode is impossible as this stands.

flink-connector-jdbc-postgres/src/main/java/org/apache/flink/connector/jdbc/postgres/datastream/connection/PostgresConnectionProvider.java
```diff
@@ -0,0 +1,307 @@
...
+        return POOL_SIZE;
+    }
+
+    @Override
+    protected String poolName() {
+        return "postgres-splitter-pool";
+    }
+
+    @Override
+    protected void onConnectionEstablished() {
```
MartijnVisser, line 95: This runs on every `getOrEstablishConnection()`, not only on establish, so each call adds an `isValid()` round trip and a `pg_current_snapshot()` query.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
```diff
@@ -0,0 +1,451 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractConnectionProvider.class);
+
+    private static final int DEFAULT_POOL_SIZE = 4;
+    private static final int MINIMUM_POOL_SIZE = 1;
+
+    protected final ConnectionOptions jdbcOptions;
+    private int queryTimeoutSeconds;
+    private transient Driver loadedDriver;
+    private transient Connection connection;
+    private final Map<String, PreparedStatement> statementCache;
```
MartijnVisser, line 75: `PreparedStatement` is not serialisable and this is not `transient`, unlike the three below. Intentional?

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/enumerator/splitter/snapshot/DatabaseSplitterEnumerator.java
```diff
@@ -0,0 +1,217 @@
...
+        fillActiveTableSplitters();
+
+        while (!Thread.currentThread().isInterrupted()
+                && (!activeTableSplitters.isEmpty() || !pendingTableSplitters.isEmpty())) {
+            boolean producedSplits = false;
+            List<TableSplitterEnumerator> finished = new ArrayList<>();
+
+            for (TableSplitterEnumerator tableSplitter : activeTableSplitters) {
+                if (tableSplitter.isAllSplitsFinished()) {
+                    lineageQueries.addAll(tableSplitter.lineageQueries());
```
MartijnVisser, line 98: Plain `LinkedHashSet`, written here on the background thread and copied from the job manager thread.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/connection/AbstractConnectionProvider.java
```diff
@@ -0,0 +1,451 @@
...
+                            this.connection.close();
+                            LOG.info("Connection gracefully closed");
+                            return null;
+                        });
+
+        try {
+            futureClose.get(10L, TimeUnit.SECONDS);
+        } catch (ExecutionException e) {
+            throw new ConnectionException(e.getCause());
+        } catch (InterruptedException | TimeoutException e) {
```
MartijnVisser, line 421: The interrupt flag is dropped. Please restore it before falling through to `abort()`.

MartijnVisser (changes requested):
This adds 16 `@PublicEvolving` classes and a new HikariCP dependency, so it needs a FLIP first: https://cwiki.apache.org/confluence/display/FLINK/Flink+Improvement+Proposals. FLINK-40108 has no discussion on it either, so there is no agreed approach to check this against.

I went through the code anyway. Two verified correctness bugs and a threading issue are inline, and the Postgres module has no tests at all, which is how the first one got this far. CI is red on all four cells as well: spotless fails on `AbstractConnectionProvider`.

PR #211, opened 2026-07-14
[FLINK-40144]Avoid query database when only get table column

Mrart:


flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/database/catalog/AbstractJdbcCatalog.java
```diff
@@ -292,7 +292,7 @@ public CatalogBaseTable getTable(ObjectPath tablePath)
 
             PreparedStatement ps =
                     conn.prepareStatement(
-                            String.format("SELECT * FROM %s;", getSchemaTableName(tablePath)));
+                            String.format("SELECT * FROM %s where 0=1;", getSchemaTableName(tablePath)));
```
MartijnVisser: This line fails `spotless:check`, so you'll need to run `mvn spotless:apply`. Could you also write `WHERE` in upper case like the rest of the statement?

MartijnVisser (changes requested):
Connector/J executes the `SELECT *` in `getMetaData()`, which for a `GROUP BY` view scans the whole base table; with the predicate it reads nothing. The MySQL, Postgres, CrateDB and OceanBase (MySQL mode) catalog ITCases pass with this change. Your commit message should start with `[FLINK-40144][jdbc]`, and please add a PR description that says whether AI tooling was used.

PR #220, opened 2026-08-01
[FLINK-40273][docs] Update JDBC documentation for the 4.x module split and APIs

spuru9:

The JDBC connector documentation still describes the 3.x connector. FLINK-36659 removed the `flink-connector-jdbc` module and bumped the version to `4.0-SNAPSHOT`, but `docs/data/jdbc.yml` was never updated, and several features added since have gone undocumented.

As a result the docs point users at an artifact that 4.x no longer builds, and the DataStream examples do not compile against the current release.

## Brief change log

**Dependencies and artifacts**
 - Point `docs/data/jdbc.yml` at `4.0-SNAPSHOT` and list one variant per published module instead of the removed aggregate artifact.
 - Document the module split in both dependency sections, including that no self-contained SQL uber jar is published, so `lib/` needs `flink-connector-jdbc-core` plus the database artifact.

**DataStream API**
 - Rewrite the sink section for the Sink V2 builder (`JdbcSink.builder()` / `buildAtLeastOnce` / `buildExactlyOnce` with `sinkTo`). The documented `JdbcSink.sink()`, `JdbcSink.exactlyOnceSink()` and `JdbcXaSinkFunction` were removed together with the old module.
 - Lead the source section with `setSplitter` and document `PreparedSplitterEnumerator` and `SlideTimingSplitterEnumerator` (FLINK-38733), keeping the deprecated `setSql` path in its own section with the migration.
 - Fix `JdbcSourceBuilder.builder()` to `JdbcSource.builder()`.
 - Add a lineage section (FLINK-34467).

**Table API**
 - Correct `scan.partition.lower-bound` and `scan.partition.upper-bound` to `Long` (`JdbcConnectorOptions.java:96-105` declares `longType()`).
 - Document `compatible-mode` as OceanBase-only with its actual values, `mysql` and `oracle`.
 - Rewrite the catalog `base-url` / `default-database` semantics for FLINK-38851, including query-parameter passthrough.
 - Add CrateDB and OceanBase to the upsert grammar table, and note that Trino has no upsert.
 - Add PostgreSQL `JSON`/`JSONB` (FLINK-39224) and `UUID` (FLINK-38850) to the data type mapping.

Applied to both `docs/content` and `docs/content.zh`.

## Verifying this change

This change is a documentation update and is verified by review.

Checks performed while preparing it:
 - Every API name, option type and default was checked against `*/src/main/java` on `main`.
 - `fromData`, `sinkTo` and `DiscardingSink` (v2) were confirmed present in `flink-runtime` and `flink-streaming-java` 2.0.0, the `flink.version` on `main`.
 - No remaining references to `JdbcSink.sink`, `exactlyOnceSink`, `JdbcXaSinkFunction`, `JdbcSourceBuilder.builder`, `addSink` or `DiscardSink`.
 - Shortcode tags and code fences balance in all four documentation files.

One item worth a reviewer's eye: `docs/data/jdbc.yml` now declares nine `variants`, and I was not able to preview how `sql_connector_download_table` renders them. The per-module list is written out as prose above the shortcode so the section reads correctly either way, but confirmation against the rendered site would be useful.

 - Dependencies (does it add or upgrade a dependency): **no**
 - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: **no**
 - The serializers: **no**
 - The runtime per-record code paths (performance sensitive): **no**
 - Anything that affects deployment or recovery: **no**
 - The S3 file system connector: **no**

## Documentation

 - Does this pull request introduce a new feature? **no**
 - If yes, how is the feature documented? **docs** (this change updates existing documentation only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/content/docs/connectors/datastream/jdbc.md
+++ b/docs/content/docs/connectors/datastream/jdbc.md
@@ -29,9 +29,22 @@ under the License.
 This connector provides a source that read data from a JDBC database and
 provides a sink that writes data to a JDBC database.
 
-To use it, add the following dependency to your project (along with your JDBC driver):
+Since version 4.0 the connector is no longer published as a single artifact. It is split into
+`flink-connector-jdbc-core`, which contains the source and the sink, and one artifact per supported
+database, which contains the dialect and the catalog for that database. Add the artifact for the
+database you are connecting to (along with your JDBC driver); it pulls in
+`flink-connector-jdbc-core` transitively. For example, for PostgreSQL:
 
-{{< connector_artifact flink-connector-jdbc jdbc >}}
+{{< connector_artifact flink-connector-jdbc-postgres jdbc >}}
+
+The available database artifacts are `flink-connector-jdbc-mysql`, `flink-connector-jdbc-oracle`,
+`flink-connector-jdbc-postgres`, `flink-connector-jdbc-sqlserver`, `flink-connector-jdbc-cratedb`,
+`flink-connector-jdbc-db2`, `flink-connector-jdbc-trino` and `flink-connector-jdbc-oceanbase`. The
+Derby dialect ships in `flink-connector-jdbc-core`.
+
+The `JdbcSource` and `JdbcSink` themselves live in `flink-connector-jdbc-core` and work against any
+JDBC driver, so a database artifact is only required for the Table/SQL API, where the dialect is used
+to generate statements.
 
 Note that the streaming connectors are currently __NOT__ part of the binary distribution.
 See how to link with them for cluster execution [here]({{< ref "docs/dev/configuration/overview" >}}).
@@ -42,28 +55,24 @@ Please consult your database documentation on how to add the corresponding drive
 
 ### Usage
 
-{{< tabs "4ab65f13-607a-411a-8d24-e709f701cd41" >}}
-{{< tab "Java" >}}
 ```java
-JdbcSource source = JdbcSourceBuilder.builder()
+JdbcSource<OUT> source = JdbcSource.<OUT>builder()
         // Required
-        .setSql(...)
+        .setSplitter(...)
         .setResultExtractor(...)
+        .setTypeInformation(...)
+        .setDBUrl(...)
+        .setDriverName(...)
         .setUsername(...)
         .setPassword(...)
-        .setDriverName(...)
-        .setDBUrl(...)
-        .setTypeInformation(...)
-        
-         // Optional
-        .setContinuousUnBoundingSettings(...)
-        .setJdbcParameterValuesProvider(...)
+
+        // Optional
         .setDeliveryGuarantee(...)
         .setConnectionCheckTimeoutSeconds(...)
-        
+
         // The extended JDBC connection property passing
         .setConnectionProperty("key", "value")
-        
+
         // other attributes
         .setSplitReaderFetchBatchSize(...)
         .setResultSetType(...)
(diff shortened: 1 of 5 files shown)


spuru9:
@RocMarshal @och5351 Can you take a look. Just docs changes.

docs/data/jdbc.yml
```diff
@@ -16,7 +16,23 @@
 # limitations under the License.
 ################################################################################
 
-version: 3.3.0-SNAPSHOT
+version: 4.0-SNAPSHOT
```
MartijnVisser: Without `flink_compatibility` the download table renders "There is no connector (yet) available" on stable docs. Also 4.1.0 is released, so 4.0-SNAPSHOT isn't right for main.

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -72,15 +81,79 @@ JdbcSource source = JdbcSourceBuilder.builder()
...
 
+### SplitterEnumerator
+
+A `SplitterEnumerator` produces the `JdbcSourceSplit`s that the readers execute, and decides whether
+the source is bounded or continuously unbounded.
+
+#### PreparedSplitterEnumerator
+
+`PreparedSplitterEnumerator` builds bounded splits from a query template. With no parameters the
+query becomes a single split:
```
MartijnVisser: On main this gives zero splits, `SqlSplitterEnumerator` returns an empty list for an empty parameter array. #243 fixes that, so this is correct once that is merged.

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -72,15 +81,79 @@ JdbcSource source = JdbcSourceBuilder.builder()
...
+```
+
+Note that the two overloads differ only in whether the last argument is a `long` (batch size) or an
+`int` (number of batches). The same can be expressed explicitly with
+`PreparedSplitterNumericParameters`:
+
+```java
+PreparedSplitterEnumerator.of(
+        "select * from books where id between ? and ?",
+        new PreparedSplitterNumericParameters(1L, 10_000L).withBatchSize(1000L));
```
MartijnVisser: `PreparedSplitterNumericParameters` is `@Internal`, so I don't think we should point users at it.

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -198,59 +218,86 @@ public class JdbcSourceExample {
...
+                        .setDBUrl(...)
                         .setDriverName(...)
                         .setResultExtractor(resultSet ->
                             new Book(
                                 resultSet.getLong("id"),
                                 resultSet.getString("title")))
                         .build();
         env.fromSource(jdbcSource, WatermarkStrategy.noWatermarks(), "TestSource")
-                .addSink(new DiscardingSink());
+                .sinkTo(new DiscardingSink<>());
```
MartijnVisser: Can you add the import? Only the `sink.v2` `DiscardingSink` compiles with `sinkTo`.

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -198,59 +218,86 @@ public class JdbcSourceExample {
...
+// current
+JdbcSource.<TestEntry>builder()
+        .setSplitter(
+                PreparedSplitterEnumerator.of(
+                        "select * from testing_table where id >= ? and id <= ?",
+                        new Serializable[][] {{1001, 1005}, {1006, 1010}}))
+        ...
+```
+
+On this deprecated path, continuous unbounded reads are configured with
```
MartijnVisser: #242 also deprecates `setContinuousUnBoundingSettings` and `setOptionalSqlSplitEnumeratorState`. And `SlideTimingSplitterEnumerator` has no discovery interval, so it isn't a like-for-like replacement. Please reflect both.

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -198,59 +218,86 @@ public class JdbcSourceExample {
...
+together:
+
+```java
+// deprecated
+JdbcSource.<TestEntry>builder()
+        .setSql("select * from testing_table where ts >= ? and ts < ?")
+        .setContinuousUnBoundingSettings(
+                new ContinuousUnBoundingSettings(Duration.ofMillis(10L), Duration.ofSeconds(1L)))
+        .setJdbcParameterValuesProvider(
+                new JdbcSlideTimingParameterProvider(0L, 1000L, 1000L, 100L))
```
MartijnVisser: This throws, the provider requires `startMills > 0`. It came over from the old page, but can you fix it while you're at it?

docs/content/docs/connectors/datastream/jdbc.md
```diff
@@ -266,30 +313,20 @@ It then repeatedly calls a user-provided function to update that prepared statem
 (preparedStatement, someRecord) -> { ... update here the preparedStatement with values from someRecord ... }
 ```
 
+The two are passed together to `withQueryStatement`. A `JdbcQueryStatement` can also be supplied
+directly if the query itself has to be derived from the record.
```
MartijnVisser: `query()` doesn't take the record and the writer prepares the statement once, so the query can't be derived from the record.

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -38,8 +38,29 @@ The JDBC sink operate in upsert mode for exchange UPDATE/DELETE messages with th
...
+| CrateDB    | `flink-connector-jdbc-cratedb`    |
+| Db2        | `flink-connector-jdbc-db2`        |
+| Trino      | `flink-connector-jdbc-trino`      |
+| OceanBase  | `flink-connector-jdbc-oceanbase`  |
+| Derby      | `flink-connector-jdbc-core`       |
+
 {{< sql_connector_download_table "jdbc" >}}
 
+None of these artifacts is a self-contained SQL uber jar. When using the SQL Client or a session
+cluster, put both `flink-connector-jdbc-core` and the artifact for your database into the `lib/`
```
MartijnVisser: CrateDB also needs the postgres jar, and OceanBase needs both the mysql and oracle jars, since those dialects build on them. Might make sense to say so here.

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -447,9 +471,21 @@ As there is no standard syntax for upsert, the following table describes the dat
...
+        </tr>
+        <tr>
+            <td>OceanBase</td>
+            <td>Depends on <code>'compatible-mode'</code>: the MySQL grammar in <code>'mysql'</code>
+                mode, the Oracle grammar in <code>'oracle'</code> mode.</td>
+        </tr>
     </tbody>
 </table>
 
+Trino does not support upsert. A Trino table always operates in append mode, even when a primary key
```
MartijnVisser: Trino has no primary keys, but the sink uses the one in the Flink DDL. With that set, `JdbcOutputFormatBuilder` does an exists check plus UPDATE or INSERT and emits deletes, so it's not append mode. Whether that works depends on the Trino connector supporting UPDATE and DELETE, and no enabled test covers it.

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -475,14 +511,39 @@ Please refer to [Dependencies](#dependencies) section for how to setup a JDBC co
...
+
+- `default-database` is set and `base-url` carries no database name, for example
+  `'base-url' = 'jdbc:postgresql://localhost:5432'` with `'default-database' = 'mydb'`
+- `base-url` carries the database name and `default-database` is omitted, for example
+  `'base-url' = 'jdbc:postgresql://localhost:5432/mydb'`
+- both are set and the database names match
+
+If the two disagree, or if neither carries a database name, catalog creation fails.
+
+`base-url` may also carry arbitrary driver options as query parameters, which are passed through to
```
MartijnVisser: When `base-url` has a query string, the database name has to be in `base-url` itself, `AbstractJdbcCatalog` rejects it otherwise. Can you add that?

docs/content/docs/connectors/table/jdbc.md
```diff
@@ -1030,4 +1105,7 @@ Flink supports connect to several databases which uses dialect like MySQL, Oracl
     </tbody>
 </table>
 
+PostgreSQL `JSON` and `JSONB` columns are read and written as their textual representation. `UUID`
```
MartijnVisser: Reads work, but writes go through `setString`, which Postgres rejects for `jsonb` and `uuid` columns unless the URL sets `stringtype=unspecified`. I'd either document that or drop "written".

MartijnVisser (changes requested):
Thanks for the PR. A few statements don't match the code on main, see inline. On the rendering question, the variants render fine with one row each. I'd like FLINK-40746's PR #242 and FLINK-40747's PR #243 in first, then this rebased on both.

PR #222, opened 2026-08-27
[hotfix][jdbc] Fix Derby upgrade for Java 17

skrcode:
This is a companion fix for Dependabot PR #204.

Derby 10.17.1.0 cannot run in this repository's Java 17 CI: its classes use class-file version 63 (Java 19), while Java 17 supports up to version 61. Derby also moved `EmbeddedXADataSource` out of `derby` into the separate `derbytools` artifact after 10.14.

This keeps the dependency upgrade, but selects Derby 10.16.1.1—the newest Derby release compatible with Java 17—and adds the matching test-scoped `derbytools` artifact.

Verification on Amazon Corretto 17:

- Reproduced the exact Dependabot-head failure: `org.apache.derby.jdbc` was missing during `testCompile`.
- Verified the 10.17.1.0 class-file incompatibility and the 10.16.1.1 Java 17 class-file level directly with `javap`.
- 33 focused Derby/XA unit tests passed.
- Derby table source/sink and JDBC integration paths passed; one OpenLineage native-parser assertion fails identically on the untouched baseline because the macOS native library is absent.
- Clean full-reactor installs passed for both CI Flink versions, 2.0.1 and 2.1.1 (11/11 modules each), including Checkstyle, Spotless, compilation, and test compilation.

The change is limited to the core module's test dependencies.


flink-connector-jdbc-core/pom.xml
```diff
@@ -149,7 +149,13 @@ under the License.
...
             <groupId>org.apache.derby</groupId>
             <artifactId>derby</artifactId>
-            <version>10.17.1.0</version>
+            <version>10.16.1.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.derby</groupId>
+            <artifactId>derbytools</artifactId>
+            <version>10.16.1.1</version>
```
snuyanzin: can you clarify why we need downgrade?

skrcode: Thanks for asking. The connector CI matrix runs on Java 17, but Apache lists Derby 10.17.1.0 as requiring Java 21 or higher; 10.16.1.1 is the newest release listed for Java 17 or higher (https://db.apache.org/derby/derby_downloads.html). I also checked the jars: 10.17 classes use class-file version 63, whereas Java 17 supports up to 61. Adding derbytools at 10.17 would resolve the missing org.apache.derby.jdbc package seen in #204 test compilation, but the tests would still be unable to run on Java 17. This PR therefore uses matching 10.16.1.1 derby and derbytools artifacts. derbytools supplies EmbeddedXADataSource, which DerbyMetadata imports, and both dependencies are test-scoped.

snuyanzin: ok, thanks for clarification
can you please
1. extract version into property
2. add a one line comment about jdk17/21 there?

skrcode: Done in 0788d13. Both Derby test dependencies now use the module-level derby.version property, with a one-line comment explaining that CI uses Java 17 while Derby 10.17 requires Java 21. I verified Maven resolves the property to 10.16.1.1.

snuyanzin (approved):

PR #218, opened 2026-07-28
[FLINK-40249][jdbc] Bump Flink, PostgreSQL JDBC driver, and other dependency versions

spuru9:
## What is the purpose of the change

Routine dependency freshness pass across the connector's Maven build, split into two categories: versions that are directly exercised by this connector's core functionality ("impacting"), and versions in test-scope code or paths not exercised by this connector ("non-impacting"), bumped as routine hygiene.

JIRA: https://issues.apache.org/jira/browse/FLINK-40249

## Brief change log

**Impacting**
- Bump `flink.version` from 2.0.0 to 2.1.2
- Bump `postgres.version` (JDBC driver) from 42.7.3 to 42.7.13 in the postgres module and the backward-compatibility module

**Non-impacting**
- Bump `jackson-bom.version` from 2.13.4.20221013 to 2.22.2
- Bump `assertj.version` from 3.24.2 to 3.27.7
- Bump `log4j.version` from 2.17.2 to 2.26.1
- Bump `commons-lang3` from 3.14.0 to 3.18.0
- Migrate the MySQL test driver from the deprecated `mysql:mysql-connector-java` coordinates to the maintained `com.mysql:mysql-connector-j:8.2.0`

**Not changed**
- `sqlserver.version` (mssql-jdbc) is intentionally left at 10.2.1.jre8. The only available fix requires jumping to a much newer major-line release (13.4.0.jre8), which could not be validated against real integration tests in this environment. Left for a follow-up once properly tested.

## Verifying this change

- Full reactor `mvn test-compile` succeeds against all bumped versions (re-verified after the jackson-bom/log4j revisions above).
- The standalone `flink-connector-jdbc-backward-compatibility` module (not part of the root reactor) compiles against the bumped versions.
- Confirmed no other module references the old pinned versions.

## Does this pull request potentially affect one of the following parts:

- Dependencies (does it add or upgrade a dependency): yes (upgrades existing dependencies; MySQL test driver coordinates changed from `mysql:mysql-connector-java` to `com.mysql:mysql-connector-j`)
- The public API, i.e., is any changed class annotated with `@Public(Evolving)`: no
- The serializers: no
- The runtime per-record code paths (performance sensitive): no
- Anything that affects deployment or recovery: no
- The S3 file system connector: no

## Documentation

- Does this pull request introduce a new feature? no

##### Was generative AI tooling used to co-author this PR?

- [X] Yes (Claude Code)

Generated-by: Claude Code (claude-sonnet-5)

PR #202, opened 2026-07-03
[FLINK-40061][Connectors/JDBC] Update parent version and flink.version in flink-connector-jdbc pom.xml

ocb3916:
## What is the purpose of the change

Update `flink-connector-jdbc`'s pom.xml to align with the latest
parent artifact and Flink core version:

- `flink-connector-parent` version bumped from `1.1.0` to `2.0.0`
- `flink.version` property bumped from `2.0.0` to `2.1.0`

These bumps keep the connector aligned with the latest
flink-connector-parent and Flink core releases.

## Brief change log

- Updated `flink-connector-parent` version: `1.1.0` -> `2.0.0`
- Updated `flink.version` property: `2.0.0` -> `2.1.0`

## Verifying this change

This change updates the parent POM and Flink core version references
only; no source code changes are involved. Existing build and test
suite should be run against the updated parent/Flink versions to
confirm compatibility.

## Does this pull request potentially affect one of the following parts

  - Dependencies (parent POM and Flink core version): yes
  - The public API, i.e., is any changed class annotated with
    `@Public(Evolving)`: no
  - The serializers: no
  - The runtime per-record code paths: no
  - Anything that affects deployment or recovery: no
  - The S3 file system connector: no

## Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? not applicable

PR #228, opened 2026-09-20
[FLINK-40723][ci] Test JDK 21 and current Flink minor lines in CI

spuru9:

JIRA: https://issues.apache.org/jira/browse/FLINK-40723

Routine CI-matrix currency pass, split by workflow:

- `push_pr.yml`: test PRs against JDK 17 **and 21** (the shared reusable `apache/flink-connector-shared-utils` workflow already supports a multi-JDK matrix via a comma-separated `jdk_version` string), and against the currently relevant Flink minor lines (`2.1.3`, `2.2.1`, `2.3.0`).
- `weekly.yml`: fixes `main`'s `flink_branches` entry, which was pinned to the stale `2.2-SNAPSHOT` even though `apache/flink`'s actual `master` now builds `2.4-SNAPSHOT`; `main` has no `jdk` pinned, so it floats to the default `'11, 17, 21'`. `v4.1` (both the `2.3.0` and `2.2.1` entries) is pinned to `jdk: '17'`. `v3.4` (Flink `1.20.x`) is pinned to `jdk: '8'`, matching the JDK that line originally supported, and bumped to the latest `1.20.x` patch (`1.20.5`).
- `pom.xml`: bump `flink.version` from `2.0.0` to `2.1.2` so the connector's actual declared build target matches what the CI matrix now tests, instead of trailing behind it.

## Brief change log

- Bump `push_pr.yml` `jdk` matrix from `['17']` to `['17, 21']`
- Bump `push_pr.yml` `flink` matrix from `[2.0.1, 2.1.1]` to `[2.1.3, 2.2.1, 2.3.0]`
- Fix `weekly.yml` `main` entry's Flink version: `2.2-SNAPSHOT` -> `2.4-SNAPSHOT`
- Default `weekly.yml` `jdk_version` fallback to `'11, 17, 21'` (used by `main`, which has no `jdk` pinned)
- Add a `v4.1` / `2.3.0` entry to `weekly.yml`, pinned to `jdk: '17'`, alongside the existing `v4.1` / `2.2.1` entry (also pinned to `'17'`)
- Pin `weekly.yml` `v3.4` to `jdk: '8'` and bump its Flink patch from `1.20.3` to `1.20.5`
- Bump `pom.xml` `flink.version` from `2.0.0` to `2.1.2`

## Verifying this change

- Full reactor compile (`flink-connector-jdbc-core` + dependents) succeeds against `flink.version=2.1.2`.
- `push_pr.yml` and `weekly.yml` validated as well-formed YAML.

- Dependencies (does it add or upgrade a dependency): yes (`flink.version` 2.0.0 -> 2.1.2)
- The serializers: no

## Documentation

- [X] Yes (Claude Code)

Generated-by: Claude Code (claude-sonnet-5)


snuyanzin (changes requested):
please do not use AI as a co-author

.github/workflows/weekly.yml
```diff
@@ -31,21 +31,24 @@ jobs:
...
+          flink: 2.4-SNAPSHOT,
           branch: main
         }, {
-          flink: 2.2.0,
+          flink: 2.3.0,
+          jdk: '17',
+          branch: v4.1
+        }, {
+          flink: 2.2.1,
           jdk: '17',
```
snuyanzin: why do we need jdk17 here if default is 17, 21?
i would expect no jdk field here at all, then it will be built for both 17 and 21

pom.xml
```diff
@@ -56,7 +56,7 @@ under the License.
     </modules>
 
     <properties>
-        <flink.version>2.0.0</flink.version>
+        <flink.version>2.1.2</flink.version>
         <scala.binary.version>2.12</scala.binary.version>
         <scala-library.version>2.12.7</scala-library.version>
         <jackson-bom.version>2.13.4.20221013</jackson-bom.version>
```
snuyanzin, line 62: should we bump other dependencies?

spuru9: Have a PR open for that https://github.com/apache/flink-connector-jdbc/pull/218, will check the ci error.

.github/workflows/weekly.yml
spuru9: Didnt want to run a lot of lefts given it already increases 6 to 9. Will add.

spuru9:
> please do not use AI as a co-author

Have created https://issues.apache.org/jira/browse/FLINK-40722

pom.xml
spuru9: https://github.com/apache/flink-connector-jdbc/pull/202 this has changes for the parent as well, which is open

.github/workflows/weekly.yml
snuyanzin: yep, probably need to leave only one version of jdk (probably 17 for all, even default)

in case of 1.20 it should be only 8

the whole range of jdk versions will be in weekly, not in PR ci

sorry for back and forth

spuru9:
Will look into the failures locally and push the fixes. Given there may be other changes, will create a JIRA.

.github/workflows/weekly.yml
spuru9: Will make the change. 
Moving to draft, will run some more CI checks locally over all combinations once more before opening the PR. Seems like java 21 is not fully compatible with the tests right now.

snuyanzin:
@spuru9 this should help with jdk21 dependencies in tests btw https://github.com/apache/flink-connector-jdbc/pull/230

.github/workflows/weekly.yml
spuru9: If I understand it correctly
Have moved the 1.20 to just 8. and other older 2.x are just 17.
For the latest it would be nice to test across a couple version so kept it at 11, 17, 21 (similar to kafka)
The total count remains the name (6)

Have kept PR for 2.1.x up given we will have that a min support as per the pom. so 6 runs as well.

snuyanzin: 1 jdk per Flink version

we could have different jdks however then need a clarifying comment why it is like that

spuru9: Done 1/version. Kept 21 for main, to check spanning 8 → 17 → 21.
My only concern is to avoid any flakiness that comes with 21 tests, which we can see in this weekly.

spuru9:
Rebased post #230

spuru9:
Have created FLINK-40725 to track the failure in jdbcDynamicTableSinkITCase.testReadingFromChangelogSource

spuru9:
Now push_pr running only with jdk 17, added latest 2.3.0 and removed 2.0.x from the matrix.

In the weekly, added 2.2.x with jdk 17 and 2.3.x with all jdk (11, 17, 21). This should cover all the scenarios exhaustively. Adding further versions don't add much benefit.

spuru9:
@MartijnVisser Have rebased the PR after https://github.com/apache/flink-connector-jdbc/pull/240

MartijnVisser (approved):
LGTM

PR #232, opened 2026-09-20
[FLINK-40725][tests] Fix sink/catalog ITCases for Flink 2.3.0 ON CONFLICT validation

spuru9:

Flink 2.3.0 (FLIP-558, FLINK-38926) tightened the planner: a query whose upsert key differs from the sink primary key is now rejected unless the `INSERT` declares an `ON CONFLICT` clause. Several ITCases that write a changelog / grouped result into an upsert sink therefore fail on Flink 2.3.0 with a `ValidationException`, while passing on 2.1.x/2.2.x. See FLINK-40725.

This makes those tests compatible across the supported Flink range and adds Flink 2.3.0 to the PR CI matrix so the paths are actually exercised.

## Brief change log

  - Affected tests, now fixed:
    - `JdbcDynamicTableSinkITCase.testReadingFromChangelogSource` (shared base, all database modules)
    - `MySqlCatalogTestBase.testGroupByInsert` (MySql)
    - `OceanBaseMysqlCatalogITCase.testGroupByInsert` (OceanBase)
  - Append `ON CONFLICT DO DEDUPLICATE` to those `INSERT`s only when running against Flink >= 2.3.0. `DO DEDUPLICATE` keeps the latest record per key, matching the prior upsert semantics.
  - Version detection lives in a new shared test helper, `FlinkVersionUtils` (in the core test-jar), backed by `EnvironmentInformation.getVersion()` — this compiles against every supported Flink version, unlike the `FlinkVersion` enum constants for 2.2/2.3 which do not exist in the connector's `flink.version=2.0.0` baseline.
  - Add `2.3.0` to the `push_pr.yml` `flink` matrix so these fixes are exercised in CI.

## Verifying this change

This change is already covered by existing tests.

All three affected tests were run locally against Flink `2.1.3`, `2.2.1`, and `2.3.0`, on JDK 17 and 21, and pass on every combination; the default `flink.version=2.0.0` build still compiles. The `2.3.0` job added to `push_pr.yml` exercises the fixed paths in CI.

  - Dependencies (does it add or upgrade a dependency, including `flink.version` or a database driver version): no
  - The public API, i.e., is any changed class annotated with `@Public(Evolving)` or `@Experimental`, or are the Table options or the JDBC dialect SPI changed: no
  - Checkpointed state, its serializers, or exactly-once delivery: no
  - The per-record code paths: no

## Documentation

---

- [X] Yes (Claude Code)

Generated-by: Claude Code (Claude Opus 4.8)


.github/workflows/push_pr.yml
```diff
@@ -28,7 +28,7 @@ jobs:
   compile_and_test:
     strategy:
       matrix:
-        flink: [2.0.1, 2.1.1]
+        flink: [2.0.1, 2.1.1, 2.3.0]
```
spuru9, line 31: Temporary change to test the resolution against 2.3.x. To be added in #228.

snuyanzin (changes requested):
why do we need `FlinkVersionUtils` and can't use `FlinkVersion`?

spuru9:
> why do we need FlinkVersionUtils and can't use FlinkVersion?

I've switched the helper to use the public FlinkVersion.current() instead of EnvironmentInformation.

It still needs to be a small helper rather than an inline FlinkVersion.current().isNewerVersionThan(FlinkVersion.v2_3), because the connector compiles against its baseline flink.version and that flink-annotations only declares FlinkVersion up to the baseline's own version. Referencing v2_3 fails to compile there — and this holds for the planned 2.1.2 baseline too, whose enum stops at v2_1:

```
error: cannot find symbol
  return FlinkVersion.current().isNewerVersionThan(FlinkVersion.v2_3);
                                                                ^
  symbol: variable v2_3   (flink-annotations 2.1.2 declares up to v2_1)
```

And as a inline is not possible, and its a block of code, I am adding a Util. Can add a TODO for cleanup post the compile version has moved 2.3+.

If there is another way let me know.

snuyanzin:
I still don't understand why we need to do any kind of `toString` and `split`

why can't we just use
```java
if (FlinkVersion.current().isNewerVersionThan(FlinkVersion.2_2)) {
...
} else {
...
}
```
?

spuru9:
Just to check my understanding, when we compile against `-Dflink.version=2.1.2`, we will be overriding the pom.xml arg and get link-annotations:2.1.3, which does not have a `FlinkVersion.2_3`. Wont that be causing a compile time failure for the test matrix with 1.20.x, 2.0.x etc?

snuyanzin:
that's true
however couldn't you still use just regular string comparison?
Given the fact there is a dot in version
2.2 is always < 2.3, same for 2.1, 1.20

so no need for parsing/splitting/convertion to numbers

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/table/sink/JdbcDynamicTableSinkITCase.java
```diff
@@ -362,9 +363,14 @@ protected void testReadingFromChangelogSource() throws Exception {
                                 "'sink.buffer-flush.max-rows' = '2'",
                                 "'sink.buffer-flush.interval' = '0'")));
 
+        String onConflict =
+                FlinkVersion.current().toString().compareTo("2.3") >= 0
```
snuyanzin:
```
                FlinkVersion.current().toString().compareTo("2.3") >= 0
```
probably better to extract it into helper method something like `isFlink558Enabled`

spuru9: Ok, I will just extract in a private function as a Util class seems to be overkill for 2 liner.

snuyanzin: yes, helper method in utility should be enough

flink-connector-jdbc-mysql/src/test/java/org/apache/flink/connector/jdbc/mysql/database/catalog/MySqlCatalogTestBase.java
```diff
@@ -513,12 +514,14 @@ void testSelectToInsert() throws Exception {
     @Test
     void testGroupByInsert() throws Exception {
         // Changes primary key for the next record.
+        String onConflict = isFlink558Enabled() ? " ON CONFLICT DO DEDUPLICATE" : "";
```
snuyanzin:
```suggestion
        String onConflict = isFlip558Enabled() ? " ON CONFLICT DO DEDUPLICATE" : "";
```
based on https://issues.apache.org/jira/browse/FLINK-38926

snuyanzin (approved):

PR #235, opened 2026-09-21
[FLINK-39206][jdbc] Give each split reader its own connection provider

MartijnVisser:

`JdbcSource` handed the same `JdbcConnectionProvider` to every split reader, while
flink-connector-base builds one per split fetcher and closes the reader of a finished split while
the next split's reader is already reading. A provider holds one connection and is
`@NotThreadSafe`, so the finished reader closed the running reader's result set. FLINK-39206 has
the mechanism.

## Brief change log

  - `JdbcSource` copies the connection provider per split reader
  - `JdbcSourceSplitReader.close()` closes the provider it now exclusively owns
  - `JdbcConnectionProvider` documents the per-split-reader copy

## Verifying this change

This change added tests and can be verified as follows:

  - Added `JdbcSourceTest#testSplitReadersDoNotShareAConnection`, red with the copy removed
  - Stress on `DerbyDynamicTableSourceITCase#testLimit`: base 2 failures in 200 runs, head 0 in 299
  - `mvn verify` on `flink-connector-jdbc-core`, plus the MySQL and Db2 source ITCases

  - The public API, i.e., is any changed class annotated with `@Public(Evolving)`: yes
  - The serializers: no
  - Anything that affects deployment or recovery: JobManager (and its components), Checkpointing, Kubernetes/Yarn, ZooKeeper: no

`JdbcSource` and `JdbcConnectionProvider` are `@PublicEvolving`; no signature changed and the new
`splitReaderSupplier` is package private and `@VisibleForTesting`. A user-supplied provider is now
copied per split reader, so one that wraps a pool gets a copy per reader; the interface Javadoc
says so. Peak connections per subtask go from one to two during a split hand-over, though the
total over a scan is unchanged. Note japicmp gives no signal in this repo: `japicmp.referenceVersion`
is `3.0.0-1.16`, which `flink-connector-jdbc-core` was never published at.

## Documentation

---

- [X] Yes (please specify the tool below)

Generated-by: Claude Code (Opus 5)


flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReader.java
```diff
@@ -222,9 +222,9 @@ public void wakeUp() {}
     @Override
     public void close() throws Exception {
         closeResultSetAndStatement();
-        if (connection != null && !connection.isClosed()) {
-            connection.close();
-        }
+        // The provider is this reader's own copy, so closing it here also runs the teardown of a
```
rionmonster: Couldn't we cover _both_ of these cases (e.g. handling the connection provider _and_ explicitly closing an open connection)? I'd imagine if we wrap the `closeResultSetAndStatement()` and the previous connection closure inside of a `try` and handle the provider closure in a `finally` it'd ensure everything is closed properly, even in cases of exceptions?

rionmonster (approved):

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/reader/JdbcSourceSplitReader.java
MartijnVisser: Both of those leak, I checked. If the statement close throws the provider close never runs, and a provider that doesn't track the connection it hands out never closes it either. I went with `IOUtils.closeAll` rather than try/finally, since that also keeps the original exception when both fail.

rionmonster: Thanks for the clarification @MartijnVisser! LGTM otherwise!

rionmonster (approved):

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSourceTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+import static org.apache.flink.connector.jdbc.JdbcTestFixture.INPUT_TABLE;
+import static org.apache.flink.connector.jdbc.JdbcTestFixture.TEST_DATA;
+import static org.apache.flink.connector.jdbc.JdbcTestFixture.TestEntry;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link JdbcSource}. */
+class JdbcSourceTest extends JdbcDataTestBase {
+
+    private static final String SQL = "select id, title, author, price, qty from " + INPUT_TABLE;
```
snuyanzin: nit: better to follow style
keywords should be always uppercase, then it would be easier to read

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSourceTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @SuppressWarnings("unchecked")
+    private static JdbcSourceSplitReader<TestEntry> newSplitReader(
+            Supplier<SplitReader<RecordAndOffset<TestEntry>, JdbcSourceSplit>> supplier,
+            String splitId) {
+        JdbcSourceSplitReader<TestEntry> splitReader =
+                (JdbcSourceSplitReader<TestEntry>) supplier.get();
+        splitReader.handleSplitsChanges(
+                new SplitsAddition<>(
+                        Collections.singletonList(new JdbcSourceSplit(splitId, SQL, null, null))));
```
snuyanzin:
```suggestion
                        List.of(new JdbcSourceSplit(splitId, SQL, null, null))));
```

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSource.java
```diff
@@ -120,16 +124,27 @@ public Boundedness getBoundedness() {
...
+
+    /**
+     * Supplies the split readers of one {@link JdbcSourceReader}. Every reader gets its own
+     * connection provider: a provider holds a single connection and is not thread safe, while
+     * flink-connector-base builds a split reader per split fetcher and closes the reader of a
+     * finished split while the reader of the next split is already reading. Sharing one provider
+     * lets the first close the connection out from under the second.
+     */
+    @VisibleForTesting
+    Supplier<SplitReader<RecordAndOffset<OUT>, JdbcSourceSplit>> splitReaderSupplier(
```
snuyanzin: seems we still use old ArchUnit here...
newer (1.4.x ) would require regeneration after adding annotation like this

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSourceTest.java
MartijnVisser: Done.

flink-connector-jdbc-core/src/test/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSourceTest.java
MartijnVisser: Agreed for main going forward, but this has to backport to v3.4 which still runs JDK 8 in the weekly, and the repo compiles to Java 8 bytecode anyway. Keeping `Collections.singletonList` here and opening a follow-up to raise the target on main.

flink-connector-jdbc-core/src/main/java/org/apache/flink/connector/jdbc/core/datastream/source/JdbcSource.java
MartijnVisser: The rule exempts same-class calls and `createReader`/`splitReaderSupplier` are both in `JdbcSource`, so nothing to regenerate here, architecture tests are green. The 1.2.0 comes from flink-architecture-tests-base; I'll open a follow-up to bump it.

snuyanzin (approved):

